PR #10, opened 2019-05-09
New Post: Code size #1

franc0is:
Add the first post in the "code size" series. This first post focuses on *measuring* code size, before optimizing.


_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+```terminal
+francois-mba:with-libc francois$ arm-none-eabi-size build/with-libc.elf
+   text    data     bss     dec     hex filename
+  10800     104    8272   19176    4ae8 build/with-libc.elf
+```
+
+So in this program we have:
+* 10800 bytes of `text`, which is code
+* 104 bytes of `data`, which is statically initalized memory
+* 8272 bytes of `bcc`, which is zero-initialized memory
```
chrisc11: `bss`

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+Wong](https://devblogs.microsoft.com/oldnewthing/20041216-00/?p=36973) and [Unix
+programmers](https://homepage.cs.uri.edu/~thenry/resources/unix_art/ch12s02.html)
+agree on!
+
+So before you do anything, measure where your code size is going. You may be
+surprised!
+
+### Measuring overall code size
+
+The simplest way to measure code size is to inspect the different sections of
```
chrisc11: `-Wl,--print-memory-usage` is pretty cool too. Suffers the fate of data being lumped into ram count but at link you get something like:

```
Memory region         Used Size  Region Size  %age Used
             rom:       10800 B       256 KB      4.12%
             ram:        8376 B        32 KB     25.56%
```

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+```
+
+The 104 bytes overflow hints at the cause: we are overflowing our flash by
+the size of our `data` section. This is because initialization values for our
+initialized static variables are stored in flash as well.
+
+To avoid confusion, I prefer to use a small script to compute the sizes:
+
+```bash
+#!/bin/bash
```
chrisc11: MAYBE: Might be neat to publish a python package that does this type of thing.

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+00001122 00000416 T usart_init
+00002808 00000444 T _sercom_get_default_pad
+00002216 00000468 T _sercom_get_async_baud_val
+00008384 00000532 T _printf_i
+00007544 00000620 T _vfiprintf_r
+00007544 00000620 T _vfprintf_r
+00000430 00000692 t _usart_set_config
+```
+
+Here we see that our largest symbol is `_usart_set_config` which is takes 692
```
chrisc11: `is takes` -> `takes up` or `uses` ?

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+
+`nm` allows us to print out all of the symbols in an ELF file, and with the
+`--size` option it will also print the size of each symbol, and even sort them
+by size if you specify `--size-sort`.
+
+So if we want to dig into what symbols consume the most code space, we simply
+do:
+
+```terminal
+francois-mba:with-libc francois$ arm-none-eabi-nm --print-size --size-sort --radix=d build/with-libc.elf | tail -30
```
chrisc11: MAYBE: `-l` is nice to tack on too for line info

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+enhancing assembly
+enhancing call tree
+enhancing siblings
+unmangling c++ symbols
+ * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit
+```
+
+Now open your favorite web browser at localhost:5000, and enjoy!
+
+Note: You may notice that some code size will be attributed to symbols under
```
chrisc11: Specifically the newlib bundled with gcc doesn't build with `-g` fwiw :(

_drafts/firmware-code-size-optimization-1.md
```diff
@@ -0,0 +1,302 @@
...
+Using puncover is relatively straightforward. You'll need to lightly prepare
+your codebase, setup the tool itself, and run it.
+
+### Preparing our codebase
+
+In order to for puncover to work, our elf file needs to contain the requisite
+information. As such, you'll need to make sure you've got the following CFLAGS
+set:
+
+* `-g`: this generates debug information which puncover requires for analysis
```
chrisc11: puncover doesn't need it but you get some extra goodies (like macro definitions) in elf when you use `-g3` (default level is 2)

_drafts/firmware-code-size-optimization-1.md
franc0is: 🤦‍♂

_drafts/firmware-code-size-optimization-1.md
franc0is: Added a note, though this is an inferior solution to `size` because you cannot compute your actual ROM usage from this output, since it does not break out data vs. bss

_drafts/firmware-code-size-optimization-1.md
franc0is: This would be ~1 day of work

_drafts/firmware-code-size-optimization-1.md
franc0is: Fixed

_drafts/firmware-code-size-optimization-1.md
franc0is: Added a note, but didn't copy the output as it gets quite messy.

_drafts/firmware-code-size-optimization-1.md
franc0is: Added a note!

_drafts/firmware-code-size-optimization-1.md
franc0is: Not going to add that here

PR #15, opened 2019-06-17
Add subscribe and RSS feed links to the home page

tyhoff:
- I added a "RSS Feed" and a "Subscribe" link to the header for all pages.
- I also reduced the padding between the header and content since it was
now quite far down the page until content was shown
- Bonus change to fix @ to ©


tyhoff:
Will add a form input a little later. Styling it to be something that isn't an eye-sore isn't worth the effort, since we don't have any sort of css framework included within the blog.

PR #22, opened 2019-07-10
ARMv6-M & ARMv7-M MPU Primer

chrisc11:


_drafts/armv7_mpu.md
```diff
@@ -0,0 +1,376 @@
...
+
+To reduce the number of writes needed to change the configuration, the `VALID` bit can be set to `1` and the `REGION` field can be set to change the value of `MPU_RNR` and apply the write to a different region. For example, if we wanted to set `REGION 7` to have a base address of `0x20000000`, we could write `0x20000000 | 0x1 << 4 | 0x7 == 0x20000017` to `MPU_RBAR`. If we then read `MPU_RNR`, we would see its value is now `0x7`.
+
+**Caution** You may have noticed the address only occupies 27 bits and the bottom 5 bits were repurposed. What if we want to protect an address that starts with a base of `0x20000002`? You can't! The minimum region size is actually implementation defined but is usually 32 bytes. The actual size supported can be dyanmically determined by writing `0xFFFFFFE0` to `MPU_RBAR` and reading the value back. The number of trailing zeros will reveal the minimum supported alignment and consequently the minimum region size. For example, reading a value of `0xFFFFFF80` would reveal bits 5 and 6 can't be set so the minimum size supported is actually `2^7` (128) bytes.
```
franc0is: You can remove the word "actually" from most sentences, but it's especially wordy in this one.

> The minimum region size is ~~actually~~ implementation defined, ~~but~~ it is usually 32 bytes.

> The ~~actual~~ size supported can be ...

franc0is: Typo: dyamically -> dynamically

franc0is: Is there really no better way to find out the minimum region size!?!? I'd expect "the minimum region size is documented in your MCUs application notes" or something like that rather than this weird write / read.

_drafts/armv7_mpu.md
```diff
@@ -0,0 +1,376 @@
...
+
+#### 3. Region Size
+
+Indicates the size the region spans over, where the size is `2 ^ (SIZE field + 1)`. The start is the address provided as the base address in `MPU_RBAR`. So for example if the base address is `0x20000000` and the `SIZE` field is `0b10000 == 16`, the MPU region will cover `0x20000000` to `0x20000000 + 2^(16+1) = 0x20020000`.
+
+**Caution**: The `SIZE` selected here **must** be aligned with the base address provided in `MPU_RBAR`. Otherwise, you will wind up covering a region that is different than expected since the base `ADDR` is always _naturally aligned_ with the `SIZE`. For example, if we had set `ADDR` above to `0x20001000`, the same region will still wind up being covered because `0x1000` would be masked out under the hood!
+
+#### 4. Sub-Region Disable
+
+The `SRD` (_Sub-Region Disable_) field is a bitmask that let's you turn off the MPU region for _1/8 units_ of the region selected. The `SIZE` of the region must be greatter than _256 bytes_ for this to have any effect.
```
franc0is: greatter -> greater

_drafts/armv7_mpu.md
```diff
@@ -0,0 +1,376 @@
...
+For embedded development on MCUs, we aren't always so lucky. A bad memory access from a task responsible for reading an accelerometer could corrupt a buffer being used elsewhere in the system to update a display!
+
+Fortunately, not all hope is lost. Many ARM MCUs implement an optional unit, known as the **Memory Protection Unit (MPU)**, which lets you safeguard how regions of memory are accessed.
+In this article, we will deep dive into the unit and walk through a few practical examples of how it can be used to prevent bad memory accesses and security exploits on _Cortex-M_ devices.
+
+<!-- excerpt end -->
+
+_Like Interrupt? [Subscribe](http://eepurl.com/gpRedv) to get our latest
+posts straight to your mailbox_
+
```
franc0is: This article is long enough that it could benefit from a table of contents

_drafts/armv7_mpu.md
```diff
@@ -0,0 +1,376 @@
...
+```
+(gdb) c
+Continuing.
+
+Program received signal SIGTRAP, Trace/breakpoint trap.
+MemoryManagement_Handler () at ../../../main.c:20
+20    __asm("bkpt 1");
+```
+
```
franc0is: Would be useful here to print the value of the SP in GDB to verify that we have indeed hit gone over

_drafts/armv7_mpu.md
```diff
@@ -0,0 +1,376 @@
...
+$2 = 0x92
+(gdb) p/x *(uint32_t*)0xE000ED34
+$3 = 0x2003e038
+```
+
+We see our `MMARVALID` is set so we know the fault took place when the address `0x2003e038` was accessed. This is directly above the bottom of our stack, `0x2003e000`, at the top of the 64 byte MPU region we allocated! So our deep recursive routine _is_ overflowing. Even though things _appeared_ to be working we were silently corrupting the memory below our stack. This would almost _certainly_ cause subtle issues in the future when the linker placed a variable in that area or we moved a heap to that region.
+
+### Catch Writes To Flash
+
+On ARM processors, `0x0` is a _valid_ address. `0x00000000`-`0x1FFFFFFF` is specified as a `Code region` and usually maps to internal flash. On most processors writes to this address range are is a no-op. This means a `NULL` pointer dereferences and write can go uncaught. We can try it out on our MCU by adding a function that looks something like this:
```
franc0is: write to this address range are ~~is~~ a no-op

_drafts/armv7_mpu.md
chrisc11: oh, yeah, totally can ... I think it's usually in the Cortex-M-specific reference manual. I'll add a note.

_drafts/armv7_mpu.md
chrisc11: hrm, I think it gets a little bit long 
image

There's probably a way to only have the TOC show headers out to a certain level though I'm not sure how exactly to do that. I was looking at the info [here])(https://blog.webjeda.com/jekyll-toc/)

_drafts/armv7_mpu.md
chrisc11: Depending on the config in use I think it might get a little confusing because the `$sp` at this point could be pointing to the ISR stack & not the main stack if `$psp` & `$msp` are in use

chrisc11:
Thanks for the review! I pushed the edits made as a new commit on top of the others

_drafts/armv7_mpu.md
chrisc11: Heh, so I see:

M0+ trm - "The MPU supports region sizes from 256-bytes to 4Gb, with 8-sub regions per region."
M3 trm - "The smallest permitted region size is 32B, corresponding to a SIZE value of 4."
M4 trm - nothing
M7 trm - nothing

So my guess would be M0+ is 256, M3,4,7 is always 32 but I'll just leave the note I added to just check the MCU manual :)

chrisc11:
> overriden

fixed, thanks

_drafts/armv7_mpu.md
chrisc11: figured out how to remove some of the sub-headings so I added the TOC!

PR #40, opened 2019-08-21
Add firmware code size 2 draft

franc0is:


_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
+---
+title: "Shrink your Firmware with GCC"
+description: "A walk through GCC configuration options to optimize firmware code size"
```
chrisc11: "through" -> "through of" ?

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+author: francois
+---
+
+## Introduction
+This is the second post in our Firmware Code Size Optimization series. Last time, we talked about measuring code size as a precondition to improving it. Now we’ll dive leap from planning into action!
```
chrisc11: "dive leap" -> "jump"?

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+<!-- excerpt start -->
+In this post, we will review compiler options that we can use to reduce firmware code size. We will focus on `arm-none-eabi-gcc`, the GCC compiler used for ARM-based microcontroller, though most of the compile-time flags we will cover are available in other GCC flavors as well as in Clang.
```
chrisc11: microcontroller -> microcontrollers

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+{:.no_toc}
+
+## Table of Contents
+
+<!-- prettier-ignore -->
+* auto-gen TOC:
+{:toc}
+
+## Setting the stage
+In order to get measurable improvements in code size, we need a reasonably large code-base to begin with. To that end, we’ll use example code from [ChibiOS](http://chibios.org/dokuwiki/doku.php), a free and open source RTOS. Specifically, we’ll use their FatFS+USB example for STM32F1 MCUs.
```
chrisc11: nit: 

```
ChibiOS[^1]

[...]
[^1]: [ChibiOS](http://chibios.org/dokuwiki/doku.php)
```

might be more preferable for links in general but not a big deal

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+#pragma GCC pop_options
+```
+
+Note that the `push_options` and `pop_options` pragmas are needed to make sure the rest of the file is compiled with the regular options.
+
+## Linker Garbage Collection
+Now that we’ve set the appropriate optimization level, we turn our attention to dead code elimination. Most programs contain dead code: it may come from libraries you use partially, or test functions you left in the final program. 
+
+Since the compiler operates on one file at a time, it does not have enough context to decide whether a function is dead code or not.  Consider a library which exposes a function to add numbers in an array: since the files in the library do not call that function, the compiler could think it is not used when other parts of our program rely on it.
```
chrisc11: MAYBE:

> since the files in the library do not call that function, the compiler could think it is not used when other parts of our program rely on it.

feels a little wordy but not sure if this is better:

while nothing in the library itself may call this function, the compiler has no way of knowing if other files are using it

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+So dead code elimination yields close to another 10KiB!
+
+### Preventing some symbols from being garbage collected
+In some cases, you may want some code that is not explicitly called by anything else to remain in your program. One such example is interrupt handlers: while it may look to the linker like they are not called, they the hardware will jump to those addresses and expect the code to be there.
+
+There are two ways to make sure a symbol isn’t garbage collected: with the `used` attribute, or with the `KEEP` linker script command.
+
+Here’s how the `used` attribute is used:
+```c
+void foo(void) __attribute__((used))
```
chrisc11: hrm I got the following (which probably applies to the attributes above as well)

```
Compiling file: main.c
../../../main.c:66:1: error: attributes should be specified before the declarator in a function definition
 void foo(void) __attribute__((used)) {
```

I don't think this actually works. It will prevent the compiler from culling the function but its not something the linker looks at ...

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+        KEEP(*(.vectors .vectors.*))
+  	      [...]
+    } > rom
+```
+
+
+## Link-time optimization
+While linkers do not traditionally do much optimizing, this has started to change. Nowadays, all the major compilers offer Link Time Optimization or LTO. LTO enables optimizations that are not available to the compiler as they touch multiple compilation units at once.
+
+LTO comes at the cost of compilation time. Even for simple programs the difference in build time will be noticeable, some of your colleagues may complain about it!
```
chrisc11: Maybe

LTO comes with two major draw backs:
* compilation time ...
* debugability ...

fwiw, a few other drawbacks I can think of:

* deeper stack depths due to aggressive inlining (dealt with a couple stack overflows due to this)
* still all-or-nothing afaik, you cannot LTO certain units. So if you are using a topology where memory placement of certain items is important, its hard to accomplish

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+[...]
+134218584 00000442 T strcmp
+[...]
+134235240 00000664 T chvprintf.
+[...]
+```
+
+Those two symbols alone represent 1KiB of code space!
+
+By default, `arm-none-eabi-gcc` ships with [newlib](https://sourceware.org/newlib/), a libc implementation optimized for embedded systems. While newlib is quite small, ARM recently added a slimmed down version of newlib dubbed “newlib-nano”, which does away with some C99 features and rewrites part of the library directly in thumb code to shrink its size. You can read more about newlib nano on [ARM’s website](https://community.arm.com/developer/ip-products/system/b/embedded-blog/posts/shrink-your-mcu-code-size-with-gcc-arm-embedded-4-7).
```
chrisc11: "recently added" -> "added" ... I don't believe it's that recent?

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+<!-- prettier-ignore -->
+* auto-gen TOC:
+{:toc}
+
+## Setting the stage
+In order to get measurable improvements in code size, we need a reasonably large code-base to begin with. To that end, we’ll use example code from [ChibiOS](http://chibios.org/dokuwiki/doku.php), a free and open source RTOS. Specifically, we’ll use their FatFS+USB example for STM32F1 MCUs.
+
+You can reproduce our setup simply by cloning the project:
+```
```
chrisc11: Might be worth mentioning what gcc toolchain you are using 

```
# git remote -v
origin	git@github.com:ChibiOS/ChibiOS.git (fetch)
origin	git@github.com:ChibiOS/ChibiOS.git (push)
(.venv) chrisc: ~/dev/ChibiOS/
# git log -1
commit 20742714c0099084e5d63a7c6d47020bde2e81e7 (HEAD -> master, origin/master, origin/HEAD)
Author: Giovanni Di Sirio <gdisirio@gmail.com>
Date:   Sun Aug 11 12:13:52 2019 +0000

    Removed old OSAL implementations.

    git-svn-id: svn://svn.code.sf.net/p/chibios/svn/trunk@12949 27425a3e-05d8-49a3-a47f-9c15f0e5edd8
```

hrm I'm  hitting:

```
# make USE_LINK_GC=no USE_LTO=no
Compiling fatfs_diskio.c
../../../os/various/fatfs_bindings/fatfs_diskio.c:10:10: fatal error: ff.h: No such file or directory
 #include "ff.h"
          ^~~~~~
compilation terminated.
make: *** [build/obj/fatfs_diskio.o] Error 1
```

chrisc11: figured it out ... had to do this first:

```
(.venv) chrisc: ~/dev/ChibiOS/ext/
# brew install p7zip
# 7za x fatfs-0.13c_patched.7z
```

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+Note that the `push_options` and `pop_options` pragmas are needed to make sure the rest of the file is compiled with the regular options.
+
+## Linker Garbage Collection
+Now that we’ve set the appropriate optimization level, we turn our attention to dead code elimination. Most programs contain dead code: it may come from libraries you use partially, or test functions you left in the final program. 
+
+Since the compiler operates on one file at a time, it does not have enough context to decide whether a function is dead code or not.  Consider a library which exposes a function to add numbers in an array: since the files in the library do not call that function, the compiler could think it is not used when other parts of our program rely on it.
+
+The linker on the other hand has visibility into our whole program, so this is where dead code could be identified and removed. Unfortunately, linkers do not perform optimizations by default.
```
tyhoff: maybe: so this is "the stage" where

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+To enable dead code optimization on GCC, you need two things: the compiler needs to split each function into its own linker section so the linker knows where each function is, and the linker needs to add an optimization pass to remove sections that are not called by anything.
+
+This is achieved with the `-ffunction-sections`  compile-time flag, and the `-gc-sections` link-time flag. A similar process can take place with dead data and the `-fdata-sections` flag.
```
tyhoff: no comma

_drafts/firmware-code-size-2.md
```diff
@@ -0,0 +1,303 @@
...
+
+   text	   data	    bss	    dec	    hex	filename
+  54828	    228	  98072	 153128	  25628	build/ch.elf
+Creating build/ch.list
+```
+
+So dead code elimination yields close to another 10KiB!
+
+### Preventing some symbols from being garbage collected
+In some cases, you may want some code that is not explicitly called by anything else to remain in your program. One such example is interrupt handlers: while it may look to the linker like they are not called, they the hardware will jump to those addresses and expect the code to be there.
```
tyhoff: Remove they in "they the hardware will"

_drafts/firmware-code-size-2.md
franc0is: Done

_drafts/firmware-code-size-2.md
franc0is: Added a line to instructions about 7za

_drafts/firmware-code-size-2.md
franc0is: Removed it.

PR #50, opened 2019-09-25
Remove Post list margin on left

tyhoff:
In a previous change, we added `margin-left: 1.5em` on all `li,ul`
elements. Unfortunately, the post lists use `ul` so this master change
applied to every listing of posts too and now the website looks worse on
mobile.

Here we add special formatting for post lists to remove that global (and
generally sensible) margin

### DIFF
--- a/_sass/layout.scss
+++ b/_sass/layout.scss
@@ -164,9 +164,14 @@ html, body {
 /* Posts */
 
 #posts {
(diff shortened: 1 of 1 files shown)


tyhoff:
Before: 
Screen Shot 2019-09-25 at 12 08 13 AM

After:
Screen Shot 2019-09-25 at 12 07 55 AM

PR #61, opened 2019-11-04
Asserts

tyhoff:
I have not proofread yet, and will still need to work on the story a bit, but the examples are ready and most of the dos and do not's are here as well.


tyhoff:
Things I want to touch on more:
- Asserts are better than nested layers of returning errors, especially if the code is all owned by the company
- That it's bad practice to do something like `ASSERT(pdTRUE == mutex_lock(mutex));`
- burn in more that it's better to assert than to fault

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
+---
+title: "Using Asserts in Embedded Systems"
+description:
+  "Asserts are a quick and proven way to prevent bugs in embedded systems. An
```
chrisc11: MAYBE:
> An optimal assert implementation both maximizes the context around the assert while minimizing code size and RAM usage.

->

An optimal assert implementation maximizes the context around the assert with minimal code and RAM overhead.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  optimal assert implementation both maximizes the context around the assert
+  while minimizing code size and RAM usage."
+author: tyler
+image: /img/debugging-asserts/assert.png
+---
+
+The use of asserts is one of the best ways to find bugs, unintended behavior,
+programmatic errors, and to learn when systems are no longer 100% functional and
```
chrisc11: maybe: to learn -> to catch

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+need to be reset. If instrumented correctly, an assert can give a developer
+context about when and where in the code an issue took place. Despite the
+numerous benefits, the practice of using asserts in firmware is not common or
+agreed upon.
+
+<!-- excerpt start -->
+
+By using asserts proactively in embedded systems on debug and production builds,
```
chrisc11: I'd consider emphasizing _and production_

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+developers can both prevent bugs before shipping and quickly surface and fix
+them after shipping. Proper assert handling is also the safest way to handle
+issues and undefined behavior that occur in production. In this post, we'll go
+over best practices with asserts, when to use asserts, and then come up with a
+productio ready custom assert implementation for an ARM Cortex-M device, all
```
chrisc11: productio -> production

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+a development board.
+
+## Common Usages of Asserts
+
+Every firmware has places where it is and isn't appropriate to assert. The
+examples below likely apply to most non-safety-critical systems.
+
+### When malloc shouldn't fail
+
+Dyanmic memory allocation is common in embedded systems, especially in today's
```
chrisc11: Dyanmic -> Dynamic

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+Mutex *mutex_create(void) {
+  Mutex *mutex = malloc(sizeof(Mutex));
+  ASSERT(mutex);
+  mutex_init(mutex);
+
+  return mutex;
+}
+```
+
+This malloc **should never fail**, because it would mean the system is out of
```
chrisc11: Maybe also mention if it did fail and we carried on we'd be using a resource unprotected.

Maybe: also mention a scenario where failure is okay? i.e as a backpressure for some sort of network transport ... could be a bullet point in the "when not to assert" section

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  EventType_B,
+} EventType;
+
+void handle_event(Event *evt) {
+  ...
+  switch (evt->type) {
+    case EventType_A:
+      // something
+    case EventType_B:
+      // something
```
chrisc11: perhaps `break;` after the something's for clarity

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+}
+```
+
+This makes it pretty obvious that anything that falls through to the `default:`
+case is undefined behavior.
+
+### RTOS API functions fail
+
+Another reasonable place to assert is when you expect certain system function to
+succeed, and if they don't, the system is in a bad state. Below is a snippet
```
chrisc11: locked with -> locked after

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+- Avoid asserting on boot up sequences. This is the most common way to introduce
+  reboot loops.
+- Don't assert on operations that depend on the hardware behaving appropriately.
+  If a sensor says it will return a value between 0-100, it's probably best not
+  assert when it's above 100, because you can never trust today's cheap
+  hardware.
+- Don't assert on the contents of data read from persistent storage, unless it's
+  guaranteed to be valid. The data read from flash or a filesystem could be
+  corrupted.
+- If it is very likely the system will recover in a few moments, it might be
```
chrisc11: Maybe an example here? What would the system be recovering from?

example/debugging-asserts/main.c
```diff
@@ -0,0 +1,77 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
+#include "impl_common.h"
+
+// Return true or false
+static bool prv_rand_bool(void) {
+  return (bool)(rand() / 2);
```
chrisc11: Did you actually find this to be random? I think you need to seed it?

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+It's now time to start asserting ourselves. The examples provided below progress
+from a naive solution to the best one I've seen and used to date, with each
+becoming more expressive while also using less (precious) code space.
+
+> NOTE: All of these implementations will behave similarly if a debugger is
+> attached. The goal of the exercise is to improve upon asserts that are hit
+> while a debugger is **not** attached and all we have is a logging system.
+
+The code starts out in `prv_main`, and will call one of two functions,
+`prv_one_true`, or `prv_one_false`, depending on a randomly generated boolean
```
chrisc11: nit: I don't quite understand what the names mean, but not a big deal. I think prv_assert_path_one, prv_assert_path_two, etc may have made more sense to me

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+The first initial approach is to already use what exists with the standard
+library, `assert(expr)`
+
+```c
+void two_true(void) {
+  assert(0);
+}
+```
+
+With the `stdlib.h` implementation of `assert.h`, nothing is recorded for us,
```
chrisc11: hrm for newlib libc this should bottom out to something like:

```
#ifdef NDEBUG           /* required by ANSI standard */
# define assert(__e) ((void)0)
#else
# define assert(__e) ((__e) ? (void)0 : __assert_func (__FILE__, __LINE__, \
                                                       __ASSERT_FUNC, #__e))
```

Up to the implementation to implement `__assert_func` though

Pretty massive codesize hit if you keep all the args because its the file, line, function, and the assert statement itself as a string!

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+- Each assert consume a fixed number of bytes, no matter the filepath, filename,
+  or number of asserts per file.
+- By capturing both the PC and LR, we can get an extra frame in our 'backtrace'.
+- If we really wanted to micro-optimize, we can choose not to record the LR and
+  save a couple more bytes.
+
+Below we define two macros, `GET_LR()` and `GET_PC()`, to retrieve the registers
+`pc` and `lr` respectively.
+
+```c
```
chrisc11: it's weird how jekyll is rendering these all in one color but maybe it just doesn't do any fancy coloring with macros

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+#define GET_LR() __builtin_return_address(0)
+#define GET_PC(_a) __asm volatile ("mov %0, pc" : "=r" (_a))
```
chrisc11: maybe worth noting this is arm & gnu gcc specific syntax

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+The drawback to this approach is that we'll have to use `addr2line` or a similar
+tool to go from our register values to the file and line information, but this
+can be added to a simple script and used by the entire team with ease.
+
+```
+$ arm-elf-linux-addr2line --exe build/nrf52.elf 0x2250
+/Users/tyler/dev/memfault/interrupt/example/debugging-asserts/impls/assert5.c:13
+```
+
+> NOTE: GCC and various optimization levels may choose to combine multiple
```
chrisc11: maybe: Compiler optimizations such as function inlining may result in _different_ asserts being merged together.

Aside: Though I'm now wondering if this winds up being try if the inline asm like we have above for the PC winds up being inserted. Not sure the compiler will merge inline asm calls

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+We can use `addr2line` on both the PC and LR to find the file and line number of
+the assert hit, as well as caller's file and line number.
+
+### Code Size Differences
+
+Why did we compile in FreeRTOS to our examples? Because we can now check out the
+differences in code sizes between our different assert implementations by
+hooking up our `MY_ASSERT` macro to the FreeRTOS `configASSERT` macro.
+
+Below are the results for compiling with `-O0` or `-Os` while only keeping in
```
chrisc11: Doesn't hurt but not sure reporting O0 matters much

Could also add some sort of delta column or overhead column

For extra effect you could wire up the default __assert_func impl with all the strings

Love the table!

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+the FreeRTOS asserts (removing those from `two_true` and `two_false`)
+
+| Assert Impl # | `-O0` CFLAG (B) | `-Os` CFLAG (B) |
+| ------------- | --------------- | --------------- |
+| None          | 20798           | 17961           |
+| 3             | 22316           | 19330           |
```
chrisc11: Could backlink to the assert # examples if you want

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+`__FILE__` assert implementations used by the common embedded platforms by using
+the `__FILENAME__` approach, or by capturing the PC/LR registers on assert.
+
+## Closing
+
+We as embedded developers should be using every tool in the toolbox to help us
+and our teams ship rock solid code, firmware, and embedded products. Asserts are
+only one piece of this puzzle, and they need to be paired with solid debugging
+infrastructure such as logging, postmortem backtrace and coredump collection,
+and automted analysis of these diagnostics.
```
chrisc11: automted -> automated

example/debugging-asserts/impls/assert2.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <assert.h>
```
chrisc11: don't need this include

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+### Article Disclaimers
+
+- This post is about runtime asserting, not compile time asserting, such as
+  [`static_assert`](https://en.wikichip.org/wiki/c/assert.h/static_assert).
+- This post is targeting the ever-growing number of **non-safety critical**
+  embedded systems, such as consumer electronics, Class 1 medical devices, home
+  automation devices, etc.
+- The above types of embedded systems are primarily constrained by memory and
+  code space limitations, so the information below takes this into account.
+- The examples below are implemented for the ARM Cortex-M series of MCU's. Many
```
mattkelly: I think the plural of MCUs is just `MCUs` with no apostrophe. [Wikipedia agrees](https://en.wikipedia.org/wiki/Microcontroller), at least.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+
+In my opinion, disabling asserts for any reason other than to check how much
+codesize asserts take up in the system is more than likely a **reckless**
+endeavor, especially if they are used correctly.
+
+### Article Disclaimers
+
+- This post is about runtime asserting, not compile time asserting, such as
+  [`static_assert`](https://en.wikichip.org/wiki/c/assert.h/static_assert).
+- This post is targeting the ever-growing number of **non-safety critical**
```
mattkelly: This should be consistent with `non-safety-critical` (two dashes) used later on. I think two dashes makes sense.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  of the concepts can be generalized to other architectures.
+
+## Making the Most of Asserts
+
+To get the most value from asserting, it is important to set up some postmortem
+infrastructure. When an assert takes place (or even a system fault), there is _a
+lot_ of context and data one can capture to help debug these issues without a
+debugger attached to the device. These include but are not limited to:
+
+- the file and line number the assert took place,
```
mattkelly: Consistency with punctuation or lack thereof at the end of bullet points

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+
+## Table of Contents
+
+<!-- prettier-ignore -->
+* auto-gen TOC:
+{:toc}
+
+## Asserts in Embedded Systems
+
+The use of asserts is controversial in the embedded space, **especially** in
```
franc0is: I would remove the `Asserts in Embedded Systems` section entirely. It reads like a second introduction and doesn't add meaningful information. 

Instead, add a note (one of those grey boxes) about "why you shouldn't compile asserts out in production builds".

It could say something like this:

> **Why you shouldn't compile asserts out in production builds**
> Many of the popular embedded platforms have options to compile out error handling and assertions. In our opinion this is a mistake. In the event of inconsistent behavior or memory corruption, crashing the device is often the safest thing you could do. Embedded systems can often reboot quickly, getting the system back to a known good state. Over time, it will also be noticed and provide valuable feedback back to your engineering teams. The alternative is worse: the system could behave in unpredictable way and perform poorly, lose customer information, ...etc.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+## Common Usages of Asserts
+
+Every firmware has places where it is and isn't appropriate to assert. The
+examples below likely apply to most non-safety-critical systems.
+
+### When malloc shouldn't fail
+
+Dynamic memory allocation is common in embedded systems, especially in today's
+Internet connected consumer electronics. These devices need to be able to
+receive an influx Bluetooth packets, use large HTTP buffers, and possibly load
```
mattkelly: `influx of`

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+[error](https://github.com/ARMmbed/mbed-os/blob/master/platform/mbed_error.h#L191-L202)
+[handling](https://github.com/zephyrproject-rtos/zephyr/blob/master/include/sys/__assert.h#L34-L89).
+
+In my opinion, disabling asserts for any reason other than to check how much
+codesize asserts take up in the system is more than likely a **reckless**
+endeavor, especially if they are used correctly.
+
+### Article Disclaimers
+
+- This post is about runtime asserting, not compile time asserting, such as
```
franc0is: Compile time asserts would make for a short and good subsection, and they don't justify an article on their own. Just a thought.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  code space limitations, so the information below takes this into account.
+- The examples below are implemented for the ARM Cortex-M series of MCU's. Many
+  of the concepts can be generalized to other architectures.
+
+## Making the Most of Asserts
+
+To get the most value from asserting, it is important to set up some postmortem
+infrastructure. When an assert takes place (or even a system fault), there is _a
+lot_ of context and data one can capture to help debug these issues without a
+debugger attached to the device. These include but are not limited to:
```
franc0is: remove `without a debugger attached to the device`. It's oddly specific.

_drafts/assert.md
franc0is: the file and line number the assert took place ~,~ **at**

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+
+With all of the above, especially coredumps, it becomes advantageous to add as
+many asserts to the system as possible, especially for capturing those 1 in
+10,000 hour bugs. This allows developers to quickly track down and investigate
+bugs affecting users in the field, all without having to reproduce the issue on
+a development board.
+
+## Common Usages of Asserts
+
+Every firmware has places where it is and isn't appropriate to assert. The
```
franc0is: This is a weird sentence, and reading it I'd have expected more info on when you should *not* assert. Instead the "when not to assert" is a separate section. Change the copy to say:

> Many assertions are application-specific, but we've got a few rules of thumbs on when you should use assertions. Below are some examples of when using assertions is a good idea.

_drafts/assert.md
franc0is: Remove `Bluetooth`

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+examples below likely apply to most non-safety-critical systems.
+
+### When malloc shouldn't fail
+
+Dynamic memory allocation is common in embedded systems, especially in today's
+Internet connected consumer electronics. These devices need to be able to
+receive an influx Bluetooth packets, use large HTTP buffers, and possibly load
+assets or fonts into RAM for a user interface. Keeping around statically
+allocated buffers for all of these functions would be irresponsible, especially
```
franc0is: s/irresponsible/wasteful 

irresponsible feels overly judgy and negative.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+void read_packet(void *buf, size_t buf_len) {
+  ASSERT(buf_len >= sizeof(Packet));
+  memcpy(...)
+}
+
+```
+
+and much more. The benefits of asserting on function arguments include:
+
+- developers connected to a debugger very quickly find programmatic mistakes
```
franc0is: These three things are good reason to use asserts in general. I'd move them into a "why you should use more assertions" section earlier in the article.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+- removes the need to check for argument validity and error handling code in the
+  upper layers
+- reduces chances of system faults by catching out-of-bounds accesses and
+  invalid pointer dereferences.
+
+It's not common to see asserts on arguments in vendor and library code, since
```
franc0is: This would be better as a side note (grey box) rather than main text.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+      ASSERT(0);
+  }
+}
+```
+
+If we want to make this even more understandable for future developers who come
+across this, something like the following can also be done.
+
+```c
+#define WTF ASSERT(0)
```
mattkelly: ❤️

franc0is: Maybe(?) WTF is too crude

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  }
+}
+```
+
+This makes it pretty obvious that anything that falls through to the `default:`
+case is undefined behavior.
+
+### RTOS API functions fail
+
+Another reasonable place to assert is when you expect certain system function to
```
franc0is: s/reasonable/great

mattkelly: `functions`

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+or check sizes of buffers
+
+```c
+void read_packet(void *buf, size_t buf_len) {
+  ASSERT(buf_len >= sizeof(Packet));
+  memcpy(...)
+}
+
+```
+
```
franc0is: Add a sentence or two on why asserting is better than painstakingly checking arguments. Oftentimes, passing a NULL pointer is a bug, not something that should be recovered from gracefully.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+Loading section .interrupts, size 0x40 lma 0x0
+Loading section .text, size 0x194d lma 0x40
+Loading section .data, size 0x4 lma 0x1990
+Start address 0x40, load size 6545
+Transfer rate: 2130 KB/sec, 2181 bytes/write.
+Resetting target
+(gdb)
+```
+
+#### Assert Metadata Structure
```
franc0is: You should put this after the "logging assert" and "standard library assert" approach.

_drafts/assert.md
tyhoff: I added your network back pressure. Was a good example.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+from a naive solution to the best one I've seen and used to date, with each
+becoming more expressive while also using less (precious) code space.
+
+> NOTE: All of these implementations will behave similarly if a debugger is
+> attached. The goal of the exercise is to improve upon asserts that are hit
+> while a debugger is **not** attached and all we have is a logging system.
+
+The code starts out in `prv_main`, and will call one of two functions,
+`prv_one_true`, or `prv_one_false`, depending on a randomly generated boolean
+variable. Both of functions will then call either `two_true` or `two_false`,
```
mattkelly: `Both of the`

_drafts/assert.md
tyhoff: That makes more sense. I'll change them.

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+extern sAssertInfo g_assert_info;
+```
+
+This provides fields for the ARM program counter and link register, the file and
+line number, and the message if any provided to the assert. In our example
+assert implementations, we'll fill in the fields we can access.
+
+### Standard Library Assert (#1)
+
+The first initial approach is to already use what exists with the standard
```
mattkelly: This wording feels strange IMO. Maybe something like:

```
The initial approach is to use what already exists within the standard library ...
```

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+  strncpy(g_assert_info.msg, msg, sizeof(g_assert_info.msg));
+  __asm("bkpt 2");
+}
+
+void two_true(void) {
+  MY_ASSERT(0, "Assert in `assert2.c::two_true`");
+}
+```
+
+The drawbacks of this method is that every message is stored in the resulting
```
mattkelly: `drawback` (singular)

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+**discourage** the use of asserts, which is not ideal.
+
+#### Extracted Information
+
+We can look up the file and line number of the assert hit assuming all assert
+strings are unique.
+
+### File**path** and Line Number Assert (#3)
+
+The most common approach I've seen for assets is to include the filepath and
```
mattkelly: `asserts`

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+build, `__FILE__` is 77 bytes!
+
+```
+(gdb) p __FILE__
+$1 = "/Users/tyler/dev/memfault/interrupt/example/debugging-asserts/impls/assert3.c"
+```
+
+Thankfully, the `__FILE__` string table entry is re-used between asserts in the
+same file, so the code size bloat primarily depends on the number of files with
+asserts within them. As a rough estimate, if we have asserts in 50 different
```
mattkelly: It kinda feels like you need a transition word here since the first sentence sounds very positive and the second sounds very negative. Something like `Still, as a rough estimate, ...`? Not really sure 🤷‍♂

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+We know the file and line number of the assert hit.
+
+### Register Values Only Assert (#5)
+
+The problem with all of the above assert implementations is that our code size
+is bloated due to storing string filenames, and also that our code size can
+change depending on how we name the files in our project. This encourages poor
+naming practices, as well as produces unexpected build sizes.
+
+My favorite way to implement asserts to record both the Program Counter and the
```
mattkelly: Consistency: program counter and link register weren't previously capitalized

I think a word is also missing: `asserts is to`

_drafts/assert.md
```diff
@@ -0,0 +1,605 @@
...
+Link Register within the macro. This provides the following benefits:
+
+- Each assert consume a fixed number of bytes, no matter the filepath, filename,
```
mattkelly: `consumes`

_drafts/assert.md
tyhoff: Woo, did it. That was fun.

_drafts/assert.md
tyhoff: Thanks!

_drafts/assert.md
tyhoff: Yeah, I tried using this as well, https://carbon.now.sh/, but it doesn't color it either. Sublime does a pretty good job at it, so I used it for my screenshot.

_drafts/assert.md
tyhoff: Great point. Thanks for that edit.

_drafts/assert.md
tyhoff: Thank you!

_drafts/assert.md
tyhoff: Nah, you are right. I added:

`However, most projects have a non-trivial number of files` in between

_drafts/assert.md
tyhoff: Fixed

_drafts/assert.md
tyhoff: Removed O0 and added a delta column

_drafts/assert.md
tyhoff: Awesome! Love this. Done

_drafts/assert.md
tyhoff: We actually use this structure in all assert handlers after my changes.

tyhoff:
All comments addressed!

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+  coredump can be captured at the time of the assert, which will provide a full
+  dump of system memory and registers
+
+With all of the above, especially coredumps, it becomes advantageous to add as
+many asserts to the system as possible, especially for capturing those 1 in
+10,000 hour bugs. This allows developers to quickly track down and investigate
+bugs affecting users in the field, all without having to reproduce the issue on
+a development board.
+
+## Why You Sould Assert
```
chrisc11: should

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+- Developers connected to a debugger very quickly find programmatic mistakes and
+  invalid usage of API's.
+- Asserting function arguments removes the need to check for argument validity
+  and error handling code in upper layers, which may in turn reduce code size.
+- Good placement of asserts can reduce chances of exceptions by catching
+  out-of-bounds accesses, invalid pointer dereferences, invalid state machine
+  transitions, and nonsensical operations (like a malloc of 12MB on a 64kB MCU).
+- Asserts can help developers quickly understand assumptions made by a certain
+  piece of code (e.g. a module that takes a `uint8_t` but requires a value
+  1-64).
```
chrisc11: maybe: value be within the range of 1 to 64

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+```c
+void *my_malloc(uint32_t n) {
+  ASSERT(n < 131072);
+  ...
+}
+```
+
+All of the above examples above are **bugs** introduced by a developer and they
+likely shouldn't be handled gracefully. The alternative is to return an error
+code, (hopefully not `-1`), and burden the develaoper with finding out _exactly_
```
chrisc11: might drop the `-1` comment (since it probably merits elaboration)

develaoper -> developer

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+where in the call stack a particular argument check filed.
+
+> It's not common to see asserts on arguments in vendor and library code, since
+> they aren't in the business of crashing devices, but it is sometimes
+> configurable on mature software. One example of configurable asserting in
+> library code is the `xTaskCreateStatic` function in FreeRTOS[^4].
```
chrisc11: maybe "can be found in FreeRTOS." and in the annotation say i.e see `xTaskCreateStatic`

That pattern is all over the codebase but this reads to me like its only in one function

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+
+> In this function, `configASSERT` is used to check for non-null pointers, and
```
chrisc11: Could be joined with the other block, not sure if the newline on 185 was intentional?

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+succeed, and if they don't, the system is in a bad state. Below is a snippet
+where we assert if a mutex wasn't locked after a few seconds, which likely means
+there is a deadlock somewhere else.
+
+```c
+int rv = xSemaphoreTake(s_mutex, sec_to_ticks(5));
+ASSERT(rv == pdTRUE);
+```
+
+> WARNING: Do not place code which performs operations within an `ASSERT`, such
```
chrisc11: ❤️

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+Since the filepaths can be lengthy and waste code space, one solution is to use
+only the filename itself, e.g. `assert3.c`, instead of the filepath, e.g.
+`/Users/tyler/.../assert3.c`.
+
+This can be accomplished by adding a define called `__FILENAME__` and using this
+in the assert implementation instead. In our Makefile or similar build system,
+this is defined along with the other `CFLAGS` as follows.
+
+```
+CFLAGS += -D__FILENAME__=\"$(notdir $<)\"
```
chrisc11: I think you may need to tack this onto the rule directly, otherwise this would get expanded before you want?

_drafts/assert.md
```diff
@@ -0,0 +1,662 @@
...
+differences in code sizes between our different assert implementations by
+hooking up our `MY_ASSERT` macro to the FreeRTOS `configASSERT` macro.
+
+Below are the results for compiling with `-O0` or `-Os` while only keeping in
+the FreeRTOS asserts (removing those from `assert_path_A` and `assert_path_B`)
+
+| Assert Impl #                              | `-Os` CFLAG (B) | Delta  |
+| ------------------------------------------ | --------------- | ------ |
+| None                                       | 17711           |        |
+| [#1](#standard-library-assert-1) (Default) | 43477           | +25766 |
```
chrisc11: woah, guess there's some reallly long assert expressions

_drafts/assert.md
tyhoff: Don't believe so, as it works. But I think I should for clarity as well.

chrisc11: indeed it does, neat!

PR #69, opened 2019-12-16
Add first Rust Zero to Main post

jamesmunns:
CC @franc0is 

Code for this post [lives here](https://github.com/ferrous-systems/zero-to-main).


_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
+---
+title: "From zero to main(): Bare metal Rust"
+description: "Bringing an embedded Rust application from zero to main"
+author: james
+tags: [zero-to-main]
+---
+
+For the past thirty years or so, the choices of languages for embedded systems developers has been relatively slim. Languages like C++ and Ada have found a home in some niche areas, such as telecommunications and safety critical fields, while higher level languages like Lua, Python, and JavaScript have found a home for scripting and prototyping. But most developers working on bare metal systems have been using the same two languages as long as I can remember: Assembly and C.
```
franc0is:
```suggestion
For the past thirty years or so, the choice of languages for embedded systems developers has been relatively slim. Languages like C++ and Ada have found a home in some niche areas, such as telecommunications and safety critical fields, while higher level languages like Lua, Python, and JavaScript have found a home for scripting and prototyping. But most developers working on bare metal systems have been using the same two languages as long as I can remember: Assembly and C.
```

Either "the choice ... has" or "the choices ... have"

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+But not for no reason! Languages often make trade-offs to fit the needs of the developers working with them: an interpreter to allow for rapid iteration, a heap for ease of memory management, exceptions for simplifying control flow, etc. But these trade-offs come with a price: whether it is code size, RAM usage, low level control, power usage, latency, or determinism.
```
franc0is: Style suggestion, use "but" less.

E.g. 
> ~~But~~ These trade-offs come with a price

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+<!-- excerpt start -->
+
+Since 2015, Rust has been redefining what it means to combine the best-in-class aspects of performance, correctness, and developer convenience into one language, without compromise.
```
franc0is:
```suggestion
Since 2015, Rust has been redefining what it means to combine the best-in-class aspects of performance, correctness, and developer convenience into one language, without compromise. In this post, we bootstrap a rust environment on a cortex-M microcontroller and walk through some of the benefits of using rust on embedded.
```

franc0is: Suggestion: add a note about the goal of the post in the excerpt, as such ^

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+<!-- excerpt end -->
+
+As a compiled systems language (based on LLVM), it is also capable of reaching down to the lowest levels of embedded programming as well, without losing built-in features that feel more at home in higher level languages, like a package manager, helpful compile time diagnostics, correctness through powerful static analysis, or documentation tooling.
+
+In this series of posts, we'll show how to bootstrap a Rust environment from scratch, integrate with existing C codebases, write correct code that is enforced at compile time, and more!
```
franc0is: Could choose to use this sentence in the excerpt above

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+This post is meant as a complement to the original [Zero to Main] post on the Interrupt blog, and will elide some of the explanations of hardware level concepts. If you're new to embedded development, or haven't seen that post, go read it now!
+
+[Zero to Main]: https://interrupt.memfault.com/blog/zero-to-main-1
```
franc0is:
```suggestion
[Zero to Main]: {{ '/tag/zero-to-main' |
relative_url  }}
```

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+```rust
+use core::{
+    mem::zeroed,
+    panic::PanicInfo,
+    ptr::{read, write_volatile},
+};
+```
+
+This imports the symbols into the current context so that they can be used. Most symbols in Rust are not available in a global namespace, which helps to avoid naming collisions.
```
franc0is: A link to the documentation for the symbols are available in the `core` crate could be useful here.

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+This line defines a symbol called `__RESET_VECTOR` at static scope. The type of this symbol is `fn() -> !`, which is a pointer to a function that never returns, or that "diverges". The value of this symbol is `reset_handler`, which is the name of a function in our program. Functions are a first class items in Rust (similar to Python), so we can use the names of functions as a value that represents a function pointer.
+
+```rust
+#[no_mangle]
+```
+
+This is another attribute, like our `#![no_std]`. By starting with `#[` instead of `#![`, we can tell this is a local attribute instead of a global attribute, which means it only applies to the next item, instead of the whole module.
+
+The `#[no_mangle]` attribute tells the compiler **not** to mangle this symbol, so it will show up as `__RESET_VECTOR`, not `from_scratch::__RESET_VECTOR`. This is important, so that our linker script can understand the symbol we are providing.
```
franc0is: Add a sentence about what name mangling is, most C developers aren't familiar (C++ concept)

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+
+```rust
+#[link_section = ".vector_table.reset_vector"]
+```
+
+This is another attribute that is informing the compiler to place this symbol in the `.vector_table.reset_vector` section of our linker script, which will place it right where we need it.
```
franc0is:
```suggestion
This is another attribute that is informing the compiler to place this symbol in the `.vector_table.reset_vector` section of our linker script, which will place it right where we need it. This is similar to gcc's `__attribute__((section(...)))`.
```

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+    static __sidata: u32; // Start of .rodata section
+}
+```
+
+This section defines a number of static symbols which will be provided by our linker, namely the start and end of the sections that are important for our reset handler to know about. These symbols are defined in an `extern "C"` scope, which means two things:
+
+1. They will be provided sometime later, by another piece of code, or in this case, the linker itself
+2. They will be defined using the "C" style of ABI and naming conventions, which means they are implicitly `#[no_mangle]`
+
+Some of these symbols are also declared as `mut`, or "mutable". By default in Rust, all variables are immutable, or read-only. To make a variable mutable in Rust, you must explicitly mark it as `mut`.
```
franc0is: Maybe: add a note about `const` in C and the parallel between it and `mut`.

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
...
+#[panic_handler]
+fn panic(_info: &PanicInfo) -> ! {
+    // On a panic, loop forever
+    loop {
+        continue;
+    }
+}
+```
+
+Rust has a concept of a `panic`, which is sort of like failing an assert in C. This happens when the program has hit an unrecoverable error case, and must be stopped in some way.
```
franc0is:
```suggestion
Rust has a concept of a `panic`, which is like failing an assert in C. This happens when the program has hit an unrecoverable error case, and must be stopped in some way.
```

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: In this case, I want to link directly to the first post, rather than the series (as that one covers all the stuff I skip in this post). Would you prefer that I still link to the series? Or use some kind of template to link to that post?

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: This is more looking at the difference between C, which uses non-namespaced symbols (e.g. items that show up in the linker), vs Rust (which handles namespacing directly).

I could link to the [documentation for the core library](https://doc.rust-lang.org/core/index.html), though this doesn't really directly map to symbols that appear in the final binary (as you could rename or re-export them).

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: Will do! I intended for my `__RESET_VECTOR` vs `from_scratch::__RESET_VECTOR` line to cover this, without going into details around special characters in linker symbols, or concepts like generics, but I can give a bit more context.

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: Good call!

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: My speaking style leaks into my writing a little too often :)

I'll refactor this a bit!

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,433 @@
+---
+title: "From zero to main(): Bare metal Rust"
+description: "Bringing an embedded Rust application from zero to main"
+author: james
+tags: [zero-to-main]
+---
```
jamesmunns: Do I need an image for my frontmatter?

_posts/2019-12-17-zero-to-main-rust-1.md
jamesmunns: Added a little bit more here, let me know if that fits what you were thinking of.

jamesmunns:
@franc0is back to you!

_posts/2019-12-17-zero-to-main-rust-1.md
franc0is: Image is optional, it helps show a better preview when you link is shared.

_posts/2019-12-17-zero-to-main-rust-1.md
franc0is: Got it. What you have is good.

_posts/2019-12-17-zero-to-main-rust-1.md
franc0is: 👍

franc0is:
LGTM! I'll merge it tomorrow morning and double check that it renders properly. Thanks!

_posts/2019-12-17-zero-to-main-rust-1.md
```diff
@@ -0,0 +1,437 @@
...
+```rust
+// Call user's main function
+main()
+```
+
+Since the main function we defined above is also divergent (`fn main() -> !`), we can simply call the function. If we had called a non-divergent function, we would get a compile error here!
+
+### Something just for Rust
+
+Rust does has one additional requirement for a bare metal program: You must define the panic handler.
```
chrisc11: "does has" -> "does have"

jamesmunns: Good catch! Fixed.

franc0is:
@jamesmunns Merged! The post should be live in a few minutes. It will get picked up in our email blast tonight and we'll do social media promotion in the morning.

franc0is:
Actually I'll do social media promotion now

PR #102, opened 2020-05-20
Produce a valid RSS feed

tyhoff:


franc0is:
Checked that RSS is now valid: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fdeploy-preview-102--interrupt.netlify.app%2Fblog%2Ffeed.xml

PR #108, opened 2020-06-16
Breakpoints

chrisc11:
An exploration into the internals of how breakpoints work with a focus on how they are managed with GDB & the Cortex-M Flash Patch & Breakpoint Unit.

I originally had a second half of this post almost done that walked through "Monitor Mode Debugging" & leveraging the DebugMonitor Exception to install breakpoints without any debugger. I opted to fold that out and I think I'll turn it into a post of it's own in the future.


_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+(gdb) break dummy_function_1
+Sending packet: $m80,4#35...Packet received: 2de9f341
+Sending packet: $m240,40#93...Packet received:[...]
+Sending packet: $m248,4#6b...Packet received: 2f730000
+Sending packet: $m248,4#6b...Packet received: 2f730000
+Sending packet: $m240,2#61...Packet received: 0149
+Breakpoint 1 at 0x240: file stub_functions.c, line 4.
+```
+
+We see some [`m` packets](https://sourceware.org/gdb/onlinedocs/gdb/Packets.html#index-m-packet) (`m addr,length`). These are used for requesting reads of memory but no breakpoint commands?! Let's `continue` and see what happens:
```
franc0is, line 166: Would be nice to add a sentence about why those reads happen.

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+
+`kind` is architecture specific but for ARM it indicates what instruction to use. (The instruction changes depending on whether the MCU uses the ARM or Thumb instruction set).[^9]
+
+> NOTE: The use of "software" breakpoint in GDB RSP is a bit of a misnomer. The gdb server can choose how to best install the breakpoint. For example, the server may opt to install a "software" breakpoint as a hardware breakpoint if there are still ones available.
+
+### Debugging GDB Remote Serial Protocol
+
+With that out of the way, let's take a look at the actual **GDB RSP** commands that get sent over the wire when we install a breakpoint. Conveniently GDB has a builtin debug command, `set debug remote 1`, which can be used to enable printing of every single command sent between the client and server!
+
+To facilitate breakpoint testing, I've created some test functions in the demo app `dummy_function_*` that we can install breakpoints on. We can use the `dump_dummy_funcs` CLI command to print information about them all:
```
franc0is, line 137:
```suggestion
To facilitate breakpoint testing, I've added `dummy_function_1` through `dummy_function_9` to my program so that we can install breakpoints on. I also added `dump_dummy_funcs` to the firmware CLI to print information about them all:
```

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+dummy_function_3: Starts at 0x260. First Instruction = 0x48024901
+dummy_function_4: Starts at 0x270. First Instruction = 0x48024901
+dummy_function_5: Starts at 0x280. First Instruction = 0x48024901
+dummy_function_6: Starts at 0x290. First Instruction = 0x48024901
+dummy_function_7: Starts at 0x2a0. First Instruction = 0x48024901
+dummy_function_8: Starts at 0x2b0. First Instruction = 0x48024901
+dummy_function_9: Starts at 0x2c0. First Instruction = 0x48024901
+dummy_function_ram: Starts at 0x20000000. First Instruction = 0x48024901
+```
+
```
franc0is, line 152: Add a sentence about why you dump the address & first instruction, and maybe a code snippet (or link to code) that shows these are no-ops? In case someone wants to repro this.

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+Packet vCont (verbose-resume) is NOT supported
+Sending packet: $Hc0#db...Packet received: OK
+Sending packet: $c#63...
+```
+
+We see a `$Z0,240,2` request issued which means "install a software breakpoint at address 0x240". This matches the address of `dummy_function_1` so that's what we'd expect! It looks like breakpoints are only installed when we resume.
+
+#### GDB RSP when a Breakpoint Triggers
+
+Now let's actually cause the breakpoint to be triggered by executing the function we installed the breakpoint on. This can be done by running the `call_dummy_funcs` CLI command:
```
franc0is, line 185: What does this do? Call them all in a row?

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+
+```
+shell> call_dummy_funcs
+[... GDB should have halted ...]
+```
+
+```bash
+Sending packet: $c#63...Packet received: T05hwbreak:;thread:0000DEAD;
+Sending packet: $g#67...Packet received:[...]
+Sending packet: $qfThreadInfo#bb...Packet received: m0000dead
```
franc0is, line 195: A note on this `0000DEAD` address would be interesting

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+##### Version 1 Layout
+
+![](/img/breakpoint/fp-comp-rev1.png)
+
+`ENABLE` flips on the individual comparator and `REPLACE` controls the behavior as follows:
+
+| Replace | Behavior                                                                         |
+| ------- | -------------------------------------------------------------------------------- |
+| 0b00    | N/A - Used to enable Flash Patching                                              |
+| 0b01    | Install breakpoint on instruction at `'000':COMP:'00'`.                          |
```
franc0is, line 274: I don't think this was super explicit that this means:
`0x000XXXXXXXXXXX00` where `X`s are taken from the value of `COMP`. Took me a second to understand.

franc0is: What a dumb design... Can you think of a reason why they would have done it this way?

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+
+##### Version 2 Layout
+
+The Version 2 Layout is a little more straightforward. Bit 0, Breakpoint Enable (**BE**), controls whether or not the breakpoint is enabled:
+
+![](/img/breakpoint/fp-comp-rev2-dcba.png)
+
+Bit 31:1 is the address to break on:
+
+![](/img/breakpoint/fp-comp-rev2-bpaddr.png)
```
franc0is, line 286: Maybe a note on why this doesn't limit the address space (LSb is used to indicate thumb/ARM mode anyways and unaligned instructions are not supported).

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+dummy_function_6: Starts at 0x290. First Instruction = 0x48024901
+dummy_function_7: Starts at 0x2a0. First Instruction = 0x48024901
+dummy_function_8: Starts at 0x2b0. First Instruction = 0x48024901
+dummy_function_9: Starts at 0x2c0. First Instruction = 0x48024901
+dummy_function_ram: Starts at 0x20000000. First Instruction = 0x4802be00
+```
+
+Aha, we can see `dummy_function_1` also was updated to include a breakpoint instruction, `0xbe00`. What's interesting here is this address is in the NRF52's internal flash space. This means to inject this breakpoint the flash page the address is in must have been reprogrammed.
+
+This is in fact exactly what happened. The SEGGER JLinkGDBServer implements a feature referred to in the docs as "unlimited flash breakpoints"[^11] which will patch flash regions the J-Link Software knows how to write to with `bkpt` instructions.
```
franc0is, line 528: Could add a note that says "this means that the JLink software ships with drivers for all those chips since they are not standard".

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,607 @@
...
+
+00000240 <dummy_function_1>:
+ 240:	4901        ldr	r1, [pc, #4] ; (248 <dummy_function_1+0x8>)
+ 242:	4802        ldr	r0, [pc, #8] ; (24c <dummy_function_1+0xc>)
+ 244:	f000 b966   b.w	514 <example_log>
+ 248:	0000732f    .word	0x0000732f
+ 24c:	00007315    .word	0x00007315
+```
+
+From a debugger we could perform the `pc`-relative load into `r1` and increment the `pc` instead of re-flashing the sector with the real instruction so the hardware can execute it.
```
franc0is, line 551: As mentioned, this is not clear at all. You should expand on it as it is super cool!

_drafts/cortex-m-breakpoints.md
chrisc11: I went with `'000':FP_COMP[28:2]:'00'.`. which I think is more explicit. I got confused initially too :)

Yeah, I can't think of a great reason for the layout. I think maybe someone was trying to be extra clever with the fact that arm has 16bit & 32bit instructions but even that isn't a very strong reason.

_drafts/cortex-m-breakpoints.md
chrisc11: I added a little note about it. It is interesting hehe
image

_drafts/cortex-m-breakpoints.md
chrisc11: There's also QSPI parts with XIP they support by default too fwiw!

_drafts/cortex-m-breakpoints.md
chrisc11: tried to be more descriptive, lmk what you think!

_drafts/cortex-m-breakpoints.md
```diff
@@ -0,0 +1,677 @@
...
+for **BE** because the bottom bit is used to convey whether the ARM or Thumb instruction set is
+being used rather than encoding any address information _and_ instructions must be aligned by their width.
+
+### Examining FPB Configuration
+
+Recall that GDB will uninstall breakpoints when the system halts so this is a rare situation where we can _only_ debug the configuration on a running system!
+
+The NRF52 includes a Version 1 **FPB** so from here on out our code examples will make use of that implementation. Here's a minimal snippet of code which can be used to dump the config:
+
+##### C Code to Dump FPB Config
```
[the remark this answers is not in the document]
chrisc11: A rare situation where you have to do it on device because they are uninstalled while GDB is halted so the FPB config will be blank

PR #110, opened 2020-06-25
Update boot.c

koisto:
Moved call to usart_teardown() to after printf before image_start(). Previously the usart was being disabled before calling printf.


franc0is:
Doh! Thanks @koisto !

PR #114, opened 2020-07-07
Normalize baseurl

sirlantis:
We had set `/blog` as the `baseurl`, but it looks like we actually don't want that to be our base url for all of our pages (at least we're also expecting pages like https://interrupt.memfault.com/tags to exist). One would notice the discrepancy when running Jekyll locally (occasionally running into 404s) but in production, the Netlify redirect configuration seems to have hid the issue.

Examples:
- Both these URLs yield the same content:
  - https://interrupt.memfault.com/blog/june-2020-roundup
  - https://interrupt.memfault.com/june-2020-roundup (here the Discourse embed doesn't work)
- Both these URLs yield the same content:
  - https://interrupt.memfault.com/blog/img/author/tyler.jpg
  - https://interrupt.memfault.com/img/author/tyler.jpg

Test Plan (to verify on Netlify instance):
- [x] Blog posts are still reachable through their `/blog/...` urls
- [x] Pagination should work as expected
- [x] Links to tags (from some posts) should work as expected
- [x] Links between posts should work


sirlantis:
@franc0is You added the original URL rewrite (via `200`) in 4ac725a8c39a32cb598261e4ca371c57cab9a543 (related: 6b3c32708e918b9988119836c74db754c8767260).

Should I add two explicit redirects for the two posts that existed at that point?

franc0is:
@sirlantis Here's the original intent: I want blog posts to have permalinks. I also want interrupt.memfault.com to eventually be more than the blog (e.g. a conference? job site? ...). Hence all blog posts should be canonically accessed at /blog/xxx.

Assets should *also* be served via /blog/... though that's less important since those do not need permalinks.
.

sirlantis:
@franc0is Okay, normalized everything to be hosted under `/blog`.

Added a few helpers when writing blog posts:

- `{% img_url some-image.png %}` instead of `{{ '/img/some-image.png' | relative_url }}` (was sometimes missing `relative_url`)
- `{% tag_url some-tag %}` instead of `{{ '/tag/some-tag' | relative_url }}` (was sometimes missing `relative_url`)

tyhoff:
The preview deployment is throwing 404's btw. 

And I love the `img_url` helpers! I tried doing that this past weekend (or rather, tried creating an `author_url` but tried to implement it as an `include` instead of a `plugin`. The plugin approach looks much friendlier.

sirlantis:
@tyhoff Fixed the 404s :)

sirlantis:
Okay, fixed the remaining kinks that I found. Deploy Preview looks good to me and everything is now strictly hosted under `/blog/` and there posts are only reachable on their intended `/blog/...` permalink,

tyhoff:
Checked for any dead links and it all checks out! 🚢

PR #116, opened 2020-07-14
Fix wrong image path in BLE primer post

sophiekovalevsky:
### Contribution description
While I was reading [Ble throughput Primer](https://interrupt.memfault.com/blog/ble-throughput-primer) I noticed that there were some images missing. This PR aims to fix that.


franc0is:
Thanks @sophiekovalevsky ! Appreciate you taking the time to fix this.

cc @sirlantis I think your changes broke this?

sirlantis:
@franc0is So it seems. Weird that this happened - as I used search and replace. Maybe VsCode didn't save a file that I had open already.

Thank you again, @sophiekovalevsky.

PR #122, opened 2020-07-31
fix markup typo in 2020-06-09-firmware-shell.md

bernardfitch:
I think you missed this closing brace in the markup for a link  (thanks for the great article!)


tyhoff:
Thank you for the fix @bernardfitch!

PR #128, opened 2020-08-15
Reuse submit PR and newsletter, set color box in TOC. 

sophiekovalevsky:
### Contribution description
I would like to propose the following changes. While we are reading the articles, there is always a table of content and I thought that probably we could highlight this block in some manner, I am not so sure about the colors but it can look something like this: 

image

In this way, the embedded developer can clearly identify where the main points are summarized. 

On the other hand, I noticed that in the posts there are two content that repeat quite often which is the invite to the newsletter and the to submit a pull request. As the blog will increase the posts more and more, I think we can take those content and set them a single place where we can edit it and then place it wherever is required. This is a table that represents the changes:

|                                                    Past                                                    |              Now              |
|:----------------------------------------------------------------------------------------------------------:|:-----------------------------:|
|                         `See anything you'd like to change? Submit a pull request!`                        |  {% include submit-pr.html %} |
| `_Like Interrupt? [Subscribe](http://eepurl.com/gpRedv) to get our latest posts straight to your mailbox_` | {% include newsletter.html %} |

This changes translate into:

image

image

What do you think about it?

### DIFF
--- a/_drafts/_template.md
+++ b/_drafts/_template.md
@@ -1,13 +1,13 @@
 ---
 # This is a template for Interrupt posts. Use previous, more recent posts from the _posts/
 # directory for more inspiration and patterns.
-# 
+#
 # When submitting a post for publishing, submit a PR with the post in the _drafts/ directory
 # and it will be moved to _posts/ on the date of publish.
-# 
+#
 # e.g.
-# $ cp _drafts/_template.md _drafts/my_post.md 
-# 
+# $ cp _drafts/_template.md _drafts/my_post.md
+#
 # It will now show up in the front page when running Jekyll locally.
 
 title: Post Title
@@ -27,8 +27,7 @@ Excerpt Content
 
 Optional motivation to continue onwards
 
-_Like Interrupt? [Subscribe](http://eepurl.com/gpRedv) to get our latest posts
-straight to your mailbox_
+{% include newsletter.html %}
 
 {:.no_toc}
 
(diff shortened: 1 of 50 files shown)


sophiekovalevsky:
> I definitely think the TOC needs to be separated more, just like you are doing here. I really like using the `_includes` more often and your rework for changing the newsletter signup.
> 
> I honestly think we can do that with the table of contents as well! I've submitted a small patch here that does just this: [sophiekovalevsky@c28b8d8](https://github.com/sophiekovalevsky/interrupt/commit/c28b8d8086386728c68081d8ae9ecc429a8777e9). I also removed the `Table of Contents` header from the gray section, so that all the headers line up with each other. It looked a little funky to have a single `<h2>` element with left padding.
> 
> image
> 
I agreed on this. Great observation. 
> I'm curious on your thoughts about the above screenshot.
> 
> Overall:
> 
> * I like the boxes around the common content
I do agree on this as well since the boxes provides the sense of showing a featured content. 
> * I like the colors. I do think the gray in the table of contents is a little dark, but I think it's nice that it is a different color than the "quote" call-outs in markdown, so I'd leave it as is.
Yeap, when I was trying to figure it out which color to set, I didn't want to have the same color that was being used for the other type of content. So, I decided to come up with some option that does not go far away of the actual theme color. 
> * I really like the usage of `{% include ...`, that is a big improvement.
> * I think after removing the background and left padding from `id="table-of-contents"`, ship it!
All right, sounds good to me. 
> 
> I can put up a follow-up PR that takes your inspiration for using `include` to do that with the full table of contents block and then that should be easier for everyone to use!
I do agree on this as well, =)

Please let me know how can I help to merge this, I am not sure how to proceed if you're one commit ahead of my PR. Should I leave the PR as it is or should I pull your changes on my branch so you can merge it easily?

tyhoff:
@sophiekovalevsky I can take it from here if that works for you. I'll patch on my couple of changes, push to this PR, then merge it. 

Thank you for your continued contributions!

sophiekovalevsky:
> @sophiekovalevsky I can take it from here if that works for you. I'll patch on my couple of changes, push to this PR, then merge it.
> 
> Thank you for your continued contributions!

Got it. All right, lets make it in that way then. Thanks for the time being on reviewing this.

PR #129, opened 2020-08-23
Set acknowledge bit to 'NACK' in I2C Address Issues

ScarceSam:
While reading through the "I2C in a Nutshell" post I noticed an error in the waveform under the section titled "I2C Address Issues".

The Acknowledge-bit in the example waveform is supposed to be in the in the NACK state but is a 0, the ACK state. So I've corrected it in the waveform. I'm pretty sure NACK == 1, and ACK == 0, but I am new to I2C :D 

Thanks for the great article :)


franc0is:
Good catch @ScarceSam! Thank you for the help 🥇.

PR #130, opened 2020-08-30
Table-of-Contents include

tyhoff:
We've messed up the table of contents a few times, so let's make it an include.


tyhoff:
Going to merge this since my incoming post is based upon it.

PR #137, opened 2020-09-09
Add Slack link to header

tyhoff:


franc0is (approved):
We might want to add a CTA to the articles as well, but this is a great start!

PR #142, opened 2020-09-19
Faster Debugging with Watchpoints

sophiekovalevsky:
What a great article @chrisc11 congrats :D While reading it I found out some catchs. 

I think we can hide the build folder that it's created from example folder? I am not sure if `build` name folder is used for all entire examples. What do you think? 

Aside of that, it looks like an image wasn't pushed to the repo?

On the top of that, I've enable compressed option so instead of loading the css files with comments and in that way we can speed up load process on the site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 _site
 vendor
+example/*/build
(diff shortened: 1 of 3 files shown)


memfault-phabricator (approved):
Thanks for the patches @sophiekovalevsky, looks great!

chrisc11 (approved):

chrisc11:
Thanks @sophiekovalevsky, looks great! I merged the changes

PR #143, opened 2020-09-27
Dockerize interrupt

sophiekovalevsky:
### Contribution description

One of the great features that come with Interrupt is the fact that there are different authors that share their experiences along a diverse set of embedded topics. I think that to write an article, the embedded developer should clone an instance of interrupt blog and set up a jekyll local server. This imply to install all required packages on local machine but wouldn't be great if we can have a way to support a dockerize interrupt version so that embedded developers, if they already have docker installed, can take advantage of this and start creating their article drafts over the container? 

For that I've created a draft PR that contain the changes to be made on repo in order to build an interrupt docker image and write the drafts article over the container.

It would be great to read your feedback on this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
-FROM ruby:2.7
+FROM ruby:2.7-alpine
+
+RUN apk --update add --no-cache --virtual build-dependencies build-base bash
 
 WORKDIR /memfault/interrupt
 
-ENTRYPOINT ["./interrupt-cli"]
+ENTRYPOINT ["./entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM ruby:2.7
+
+WORKDIR /memfault/interrupt
+
+ENTRYPOINT ["./interrupt-cli"]
--- a/interrupt-cli
+++ b/interrupt-cli
@@ -1,0 +1,88 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+function help()
+{
+  echo "\
+Usage: interrupt-cli [--build] [--run]
+  Build, and run interrupt blog
+
+Options:
+  -b, --build           Build and install required packages.
+  -r, --run             Launch blog at http://localhost:4000.
+  -h, --help            Display this help and exit.
+"
+}
+
+OPTSTR=brh
+LONGOPTS=build,run,help
+
+BUILD=false
+RUN=false
+
+! PARAMS=$(getopt --options=$OPTSTR --longoptions=$LONGOPTS --name "$0"  -- "$@")
+
+if [ ${PIPESTATUS[0]} -ne 0 ]; then
+    exit 2
+fi
+
+eval set -- "$PARAMS"
+
+while true; do
+  case "$1" in
+    -h|--help)
+      shift
+      help
+      exit 0
+      ;;
+    -b|--build)
+      BUILD=true
+      shift 1
+      ;;
+    -r|--run)
+      RUN=true
+      shift 1
+      ;;
+    --)
+      shift
+      break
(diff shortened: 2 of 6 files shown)


franc0is:
I like it. @tyhoff @sirlantis any opinions? A small preference of our would be to use `invoke` rather than bash scripts (see the Interrupt article about it!).

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ruby:2.7
+
+WORKDIR /memfault/interrupt
+
+ENTRYPOINT ["./interrupt-cli"]
```
sirlantis, line 5: nit: Maybe rename it to `./entrypoint.sh` to hint at that this is mostly intended to be run inside the Docker container?

interrupt-cli
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
```
sirlantis, line 4: nit: combine into somewhat idiomatic "strict bash" oneliner?

```suggestion
set -euo pipefail
```

interrupt-cli
```diff
@@ -0,0 +1,88 @@
...
+
+OPTSTR=brh
+LONGOPTS=build,run,help
+
+BUILD=false
+RUN=false
+
+! PARAMS=$(getopt --options=$OPTSTR --longoptions=$LONGOPTS --name "$0"  -- "$@")
+
+if [ ${PIPESTATUS[0]} -ne 0 ]; then
```
sirlantis, line 27: Double quote to prevent globbing and word splitting.
```suggestion
if [ "${PIPESTATUS[0]}" -ne 0 ]; then
```

interrupt-cli
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+
+if [ $RUN = true ]; then
+  concatenate_if_required
+  LAUNCH_INTERRUPT+="RUBYOPT='-W0' bundle exec jekyll serve -D"
+fi
+
+echo $LAUNCH_INTERRUPT
+eval $LAUNCH_INTERRUPT
```
sirlantis, line 88: Double quote to prevent globbing and word splitting as well as respecting quotes and whitespace.

```suggestion
echo "$LAUNCH_INTERRUPT"
eval "$LAUNCH_INTERRUPT"
```

sirlantis (commented):
Generally looks good. Shell scripts are [less than 100 lines, so "okay" to keep as is](https://google.github.io/styleguide/shellguide.html#when-to-use-shell).

sirlantis:
> I like it. @tyhoff @sirlantis any opinions? A small preference of our would be to use `invoke` rather than bash scripts (see the Interrupt article about it!).

Not only would `invoke` add the additional pain of installing and dealing with Python and the dependencies, having that inside a Ruby project instead of just using Ruby's `rake` (which is one of the inspirations for Invoke/Fabric) would be an odd choice.

That said, either would defeat the intention of this PR to provide a way to run `interrupt` without having to deal with the pain of installing Ruby (or Python) environments. So, as long as the bash scripts are small like this, I wouldn't mind the bash scripts.

franc0is:
> That said, either would defeat the intention of this PR to provide a way to run interrupt without having to deal with the pain of installing Ruby (or Python) environments. So, as long as the bash scripts are small like this, I wouldn't mind the bash scripts.

makes sense, thanks for chiming in!

sophiekovalevsky:
@franc0is, @sirlantis thank you both for the feedback provided. 

I pushed the changes. 

On the top of that, instead of using ruby image, I decided to use alpine instead. The only trade-off is that when building, it will take a little bit of extra time, but we win in shrinking the image down to ~ 257 MB.

franc0is (approved):
LGTM, thanks @sophiekovalevsky !

PR #154, opened 2020-11-11
Possible minor corrections

NevynUK:
* I think "just saw" should be "just so"
* "example" should be "examples"
* Possible incorrect unicode character


tyhoff:
Thanks for pointing these out! It seems the commit you pushed is @franc0is 's original commit, so I'll just take the fixes you mentioned in the message of this PR and push those.

PR #159, opened 2020-11-21
fix libc symbol not defined compile error

cheng3100:
Hi, I find this blog by accident and learn a lot from here. When I try to compile the example demo of `breakpoint ` and `debugmon` I get a error message at link stage.The error information is here:

```
Linking library
/usr/lib/gcc/arm-none-eabi/9.2.1/../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/9.2.1/../../../arm-none-eabi/lib/thumb/v7e-m+fp/hard/libc.a(lib_a-abort.o): in function `abort':
/build/newlib-CVVEyx/newlib-3.3.0/build/arm-none-eabi/thumb/v7e-m+fp/hard/newlib/libc/stdlib/../../../../../../../../newlib/libc/stdlib/abort.c:59: undefined reference to `_exit'

```
My platform information:
```
kernel version: 5.8.0-28
gcc version: arm-none-eabi-gcc(15:9-2019-q4-0ubuntu1) 9.2.1 20191025(release)
```

After some search I found that is because these demo use some function from libc by include head file like `stdio.h`, and the lack symbol like `_exit` is system call which the function in libc invoke. But for these demo there are no need to use a operating system so there are no system call at all. I found the best way to fix it is add a `--sepecs=nano.specs` at the link flag and it work at my platform. I also make some change to Makefile to let them more useful.

### DIFF
--- a/example/breakpoint/Makefile
+++ b/example/breakpoint/Makefile
@@ -1,10 +1,18 @@
 ROOT_DIR := $(abspath .)
 BUILD_DIR  := $(ROOT_DIR)/build
-Q ?= @
+
+ifeq ($(V), 1)
+Q =
+else
+Q = @
+endif
 
 FREERTOS_ROOT_DIR := $(ROOT_DIR)/freertos_kernel
 
-COMPILER ?= arm-none-eabi-gcc
+COMPILER ?= arm-none-eabi-
+
+CC = $(COMPILER)gcc
+DUMP = $(COMPILER)objdump
 
 PROJECT_SRC_DIR = $(ROOT_DIR)/src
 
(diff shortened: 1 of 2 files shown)


cheng3100:
fix #159

example/breakpoint/Makefile
```diff
@@ -1,10 +1,18 @@
 ROOT_DIR := $(abspath .)
 BUILD_DIR  := $(ROOT_DIR)/build
-Q ?= @
+
+ifeq ($(V), 1)
+Q =
+else
+Q = @
+endif
```
tyhoff, line 8: <3

tyhoff (approved):

PR #39, opened 2019-08-20
WIPWIPWIP: Invoke post

tyhoff:
Placing this hear just for skimming to see if I'm heading in the right direction

PR #160, opened 2020-12-02
Interrupt Repo Updates #39

bkmxer:
- Add social sharing icons t
- If possible, add a search bar on the right-side column.
- Add the subscribe to the blog form on the rightside column.


collhar (approved):

franc0is:
Would it be possible to hide the search & subscribe by default when in a narrow view (e.g. mobile)? Right now it takes up the whole screen. Ideally it would be minimized and could be expanded by clicking an icon / button.

franc0is:
Screenshot:
image

franc0is:
For the post view, I think we should add the "search" box, but leave the subscribe.
The sidebar should follow on scroll, so it's always visible while reading an article
I think the social share buttons should be on the sidebar while reading, so they're always visible (what do you think @collhar) ?

bkmxer:
> Would it be possible to hide the search & subscribe by default when in a narrow view (e.g. mobile)? Right now it takes up the whole screen. Ideally it would be minimized and could be expanded by clicking an icon / button.

Yes, we can do that,
Do we need some buttons to trigger the show/hide of search and subscribe sections?

bkmxer:
> For the post view, I think we should remove the "search" box, but leave the subscribe.
> The sidebar should follow on scroll, so it's always visible while reading an article
> I think the social share buttons should be on the sidebar while reading, so they're always visible (what do you think @collhar) ?

@franc0is 
1) For the post view, should we remove the search for all the resolutions or only for mobile?
2) yep, i can do that
3) I think that was the initial idea, but during the discussion on the call, @collhar proposed to make it at the bottom.
I can move em if needed :)

collhar:
> > For the post view, I think we should remove the "search" box, but leave the subscribe.
> > The sidebar should follow on scroll, so it's always visible while reading an article
> > I think the social share buttons should be on the sidebar while reading, so they're always visible (what do you think @collhar) ?
> 
> @franc0is
> 
> 1. For the post view, should we remove the search for all the resolutions or only for mobile?
> 2. yep, i can do that
> 3. I think that was the initial idea, but during the discussion on the call, @collhar proposed to make it at the bottom.
>    I can move em if needed :)

@bkmxer 
1. I've changed my mind and think we should keep the search bar on the right side where it is as long as it floats, but hide the full right bar on mobile for the posts and the main listing page.
2. Thanks!
3. Move the social under on the sidebar too as long as it floats.

Thanks!

bkmxer:
Hi @collhar @franc0is 
Final question, should we keep the share buttons on the main page on desktop? Or only keep them on the post page?

PR #167, opened 2020-12-16
Fix urlencoding post titles for social links

tyhoff:
Replacing `'` with nothing wasn't the right approach. Just need to URL encode the items. Unfortunately no `url_encode` built in. I think it may be in a future Jekyll version though?

### DIFF
--- a/_includes/share-buttons.html
+++ b/_includes/share-buttons.html
@@ -13,19 +13,19 @@
             </svg>
         </div>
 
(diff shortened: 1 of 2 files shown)


tyhoff:
Nope, broken as well. Darn.

tyhoff:
Fixed it

chrisc11 (approved):

tyhoff:
@collhar JSYK

PR #175, opened 2021-01-12
Balancing Test Coverage Vs Overhead

tyhoff:
Guest post by Erik Fogg from ProdPerfect.


_drafts/testing-vs-overhead.md
```diff
@@ -0,0 +1,122 @@
...
+
+*Feature criticality* is another category for test development. This involves testing features in order of how critical they are for the product to work successfully. If a feature breaking has little impact on the user experience, then it should receive fewer resources than a feature that is integral to the user experience.
+
+Two parts of the codebase that should be rigorously tested no matter are [firmware update]({% post_url 2020-06-23-device-firmware-update-cookbook%}) and factory reset. Firmware update will allow your devices to update from one firmware to another, and factory reset will bring the device back to the state it was in when it was first provisioned. Factory reset will typically involve zeroing external flash memory, deleting all user preferences, removing bonding information, wiping the filesystem, and ideally loading the original firmware that the device started with. The only things that should remain between factory resets are the device provisioning information and maybe some logs to determine what caused the factory reset in the first place.
```
chrisc11, line 86:
```suggestion
Two parts of the codebase that should be rigorously tested no matter what are [firmware update]({% post_url 2020-06-23-device-firmware-update-cookbook%}) and factory reset. Firmware update will allow your devices to update from one firmware to another, and factory reset will bring the device back to the state it was in when it was first provisioned. Factory reset will typically involve zeroing external flash memory, deleting all user preferences, removing bonding information, wiping the filesystem, and ideally loading the original firmware that the device started with. The only things that should remain between factory resets are the device provisioning information and maybe some logs to determine what caused the factory reset in the first place.
```

PR #176, opened 2021-01-19
Release Versioning

tyhoff:


_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+title: Firmware Versioning
+description:
+  Firmware versioning with relelase management in mind
+author: tyler
+image: img/<post-slug>/cover.png # 1200x630
+---
+
+Release versioning might seem like a boring topic. Honestly, it should be. There should only be a couple of right ways to do versioning, and each project should pick one of the agreed-upon methods (SemVer[^semver], CalVer[^calver], etc.) that makes the most sense to the project. We don't live in this ideal world unfortunately and many projects choose to deviate from versioning standards. I'm not here to say they are wrong, but I'd like to discuss what they might be missing out on.
+
+Along with a typical release version, it is useful to include extra metadata, such as the Git revision, [GNU Build ID]({% post_url 2019-05-29-gnu-build-id-for-firmware %}), build timestamp, and other related items. With all of this in place, both machines and developers can quickly hone in on *exactly* which release is running on a given device, what the source is, where it came from, and how to rebuild it if necessary.
```
chrisc11, line 11: > build timestamp

heh, torn on this one, from a reproducibility perspective, having timestamps in a build can be annoying because it will change a md5

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+Versions don't exist just to differentiate one release from another. They serve other purposes! Such purposes include, but are not limited to:
+
+- Providing an ordering to versions (e.g. 1.1 > 1.0)
+- Encode compatibility information, breaking changes, and upgrade/downgrade restrictions
+- Reference a specific "revision" of the software
+- Reference a specific "build" of the software
+- Tell users when *major* changes took place (e.g. Windows 8 → Windows 10)
+
+If a project chooses to use just a Git SHA, or a build timestamp, or a random ASCII name, then many or all of the above benefits that versions typically provide are lost. We can do better.
```
chrisc11, line 35:
```suggestion
If a project chooses to use just a Git SHA, or a build timestamp, or a random ASCII name, then many or all of the above benefits that versions typically provide are lost. We can do better! 
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+Normal software packages can be updated at any time, with almost no repercussions. If incompatibility arises, just revert the change and continue. However, on firmware devices, the firmware updates itself. If there is an incompatibility, it might require a re-install, a factory reset, or in the worst of cases, the device might be bricked!
+
+With firmware updates, you also can't expect every device to install every single version (unless that is a hard requirement of the update flow). If you deploy version 1.0.0, version 1.1.0, and version 1.2.0, your firmware should be able to update from 1.0.0 → 1.2.0 and skip 1.1.0 entirely. If a user puts your device in their nightstand and pulls it out 6 months later, this is exactly what should happen.
+
+However, keeping that backward compatibility can weigh down our firmware, especially if we need to keep migration code in our firmware. Migration code adds complexity, code space, and allowing installs from multiple firmware versions increases our QA test matrix dramatically.
```
chrisc11, line 69:
```suggestion
However, keeping that backward compatibility can weigh down our firmware, especially if we need to keep migration code in our firmware. Migration code adds complexity, code space, and increases our QA test matrix dramatically. 
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+This was a backward-incompatible firmware update, and I believe it warranted the MAJOR increase from 2 to 3.
+
+### Incrementing MINOR
+
+Example: 1.1.0 → 1.2.0
+
+In the context of firmware, this is likely what most of your firmware version updates will be, especially if they contain new features and functionality. A MINOR update means that your device can upgrade from a version with the same MAJOR to any *newer* version with the same MAJOR. For example, as we discussed in the MAJOR section, any firmware in the 2.x release train should be able to update to any newer 2.x version, such as 2.0.0 → 2.5.0. 
+
+A MINOR update should not allow *downgrading* firmware versions, which is going from a firmware version to an older one, such as 2.5 → 2.0. Speaking from experience, this is a nightmare to deal with, and you should just disable this behavior from ever happening. We provide a [code snippet](#detecting-downgrades) later in the article to help prevent against downgrades at the installer level.
```
chrisc11, line 85: ot: when downgrades aren't easily allowed, one challenge does become QA'ing the upgrade path and how to best do that

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Pre-release Tag Best Practices
+
+The pre-release tag is any series of dot separated identifiers consisting of ASCII letters and numbers.
+
+The most common pre-release tags are 'alpha', 'beta', and 'rc', which should satisfy most of your versioning requirements. 
+
+- Use 'alpha' when the release branch is first cut from master or another release branch. It's pretty much assumed that 'alpha' builds are unstable and should be treated as broken-by-default.
+- Use 'beta' when the API is stable, there are no longer any breaking changes and features being pushed into the release, and the stabilization phase has begun.
+- Use 'rc' when you think everything is stable and QA should start doing rigorous testing because it might turn out that this build could be "the one".
```
chrisc11, line 105: call out that rc == release candidate just to be explicit

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+>>> semver.parse_version_info('1.1.1-alpha.1') \
+    < semver.parse_version_info('1.1.1-alpha.11')
+True
+```
+
+### Missing Pieces of Semantic Versioning
+
+Semantic Versioning provides a plethora of benefits to our firmware and release management workflows, but it doesn't solve all of our requirements. One of our requirements was to be able to reference a specific build of a firmware, which SemVer does not help with.
+
+If two developers both build a 1.1.1 firmware, we now have two builds of 1.1.1 with no way to tell the difference between the two! They could be based upon two entirely different revisions of software or even built with a different compiler. We could use pre-release tags or build meta tags, but we also don't want to pollute our SemVer string with data that could be better solved by other versions.
```
chrisc11, line 125:
```suggestion
For example, if two developers locally compile a "1.1.1" firmware, we now have two builds of "1.1.1" with no way to tell the difference between the two! They could be based upon two entirely different revisions of software or even built with a different compiler. We could use pre-release tags or build meta tags, but we also don't want to pollute our SemVer string with data that could be better solved by other versions. 
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+Let's now look at Git revisions, which will help us identify the software revisions that our firmware is built from.
+
+## Git Revisions
+
+The second piece of metadata that we want to include in our firmware is a Git revision or Git tag, such as `d287bc7311`.
+
+By having the Git revision baked into the firmware, we can determine which software the binary is based upon, which would help us when we are trying to triage bugs by performing a [git bisect]({% post_url 2020-04-21-git-bisect %}) or rebuilding old firmwares (provided we have [reproducible builds]({% post_url 2019-12-11-reproducible-firmware-builds %})).
+
+You can call `git describe` to acquire the Git revision SHA with a `+` on the end if the project had a dirty state when called.
```
chrisc11, line 135:
```suggestion
You can call `git describe` to acquire the Git revision SHA with a `+` on the end if the project had a dirty state (local modifications) when called.
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+$ git describe --match ForceNone --abbrev=10 --always --dirty="+"
+8bb5d7f162+
+```
+
+The use of `--match ForceNone` is because I just want the SHA and the dirty signifier and nothing else.
+
+If you'd like to also include a tag and number of commits from the last tag like some projects have, you can use the following command:
+
+```
+# git describe --tags --always --dirty="+" --abbrev=10
```
chrisc11, line 147:
```suggestion
$ git describe --tags --always --dirty="+" --abbrev=10
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+```
```
chrisc11, line 148: maybe add an example of the output from these two commands

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+#### Git Tags & CI
+
+If you have your [continuous integration]({% post_url 2019-09-17-continuous-integration-for-firmware %}) system set up to build your firmware, pulling the SemVer from the Git tag is a great way to go. 
+
+I like to configure my CI systems to watch tags that start with `release-*`. That way, I can tag and push releases with a few simple Git commands and a script in CI which parses the SemVer and sets build variables
+
+```
+$ git checkout release-1.0
+$ git tag release-1.0.0
```
chrisc11, line 202: I think?
```suggestion
$ git tag 1.0.0
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+$ python ~/print_header.py build/fwup-example-app.bin
+Version: 1.0.1
+Git SHA: 70859a3
+```
+
+### Detecting Downgrades
+
+One of the most unpleasant and avoidable ways of bricking a device is to allow it to install incompatible versions, such as downgrades. If SemVer is followed closely, almost all dangerous downgrade paths could be avoided without much logic or defensive work.
+
+Although using `scanf` isn't really recommended due to code-size bloat, here is a simple example snippet of code to detect firmware downgrades for those that use Semantic Versioning.
```
chrisc11, line 277:
```suggestion
Although using `sscanf` isn't really recommended due to code-size bloat, here is a simple example snippet of code to detect firmware downgrades for those that use Semantic Versioning. 
```

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+
+```c
+#include <stdio.h>
+#include <string.h>
+
+int main(void) {
+    char *current_ver = "1.1.0";
+    char *new_ver = "1.0.0";
+
+    int currentMajor, currentMinor, currentPatch;
```
chrisc11, line 287: maybe zero init these in case the version is malformed
currentMajor = 0, currentMinor = 0, currentPatch = 0;

_drafts/release_versioning.md
```diff
@@ -0,0 +1,328 @@
...
+    sscanf(current_ver, "%d.%d.%d", &currentMajor, &currentMinor, &currentPatch);
+    sscanf(new_ver, "%d.%d.%d", &newMajor, &newMinor, &newPatch);
+    
+    if (newMajor < currentMajor || newMinor < currentMinor) {
+        // Dowgrade. Do not perform the "update".
+    }
+}
+```
+
+If you use another versioning scheme, you should still try to detect downgrades, whether that is by using the Git SHA, a timestamp, commit number, or anything else that is roughly stable and sequential.
```
chrisc11, line 299: Might drop SHA, not sure how that could be ordered
```suggestion
If you use another versioning scheme, you should still try to detect downgrades, whether that is by using, a timestamp, commit number, or anything else that is roughly stable and sequential.
```

chrisc11 (approved):
Great post!

_drafts/release_versioning.md
tyhoff: I'm sure you can technically generate a new timestamp if and only if something *else* changed?

_drafts/release_versioning.md
franc0is: There probably isn't a way to detect a downgrade with the Git SHA.
```suggestion
If you use another versioning scheme, you should still try to detect downgrades, whether that is by using a timestamp, commit number, or anything else that is roughly stable and sequential.
```

franc0is (approved):
Great post

PR #177, opened 2021-01-20
fix some typo's in posts

mjl-:
while reading, i found some typo's.

i also saw an unfinished sentence at https://interrupt.memfault.com/blog/how-to-write-linker-scripts-for-firmware, just below the "Variables" section: "In the first post, our ResetHandler relied on seemingly magic variables to know the address of each of our sections of memory. It turns out, those variable came" [unfinished], but don't have a suggestion for content.

thanks for writing and publishing interrupt!


tyhoff (approved):
Awesome, thanks @mjl-! Looks like we need to run Grammarly on our old posts :/

PR #178, opened 2021-01-21
Fix minor typo in 2021-01-19-release-versioning.md

bernardfitch:
Adds a missing 'C' to change a rogue 'PATH' into the 'PATCH' that I'm sure was intended

P.S. Thanks for another great article ;)


tyhoff (approved):
Thanks @bernardfitch !

PR #179, opened 2021-02-05
Update 2020-03-23-intro-to-renode.md

IvanVnucec:
added analyzer command so we can see the `hello world!` string.


franc0is (approved):
Thanks! Looks great.

PR #191, opened 2021-04-13
Improved pagination UX

MihajloNesic:
Hello 👋

I noticed that the blog page count has been increasing. Because of that, the pagination is starting to look overwhelming. Not that it looks bad at it's current state, but as the page count increases it will become overpopulated with page number buttons.

_Current_
current pagination

---

I have found a better solution for pagination. 

I took the inspiration from the following sources:
* https://dribbble.com/shots/2351417-What-s-your-fav-pagination (example number 7)
* https://uxframework.pearson.com/c/pagination (Standard pagination)
* https://codemyui.com/pagination-ux-pure-css/

Here's how it works:

The structure is as follows:
`[Previous page] [First page number] ... (N trailing) [Current page] (N leading) ... [Last page number] [Next page]`

There is a new option in **_config.yml** file called `paginate_trail` and the dafault value is `2`. It is used to generate N (in default case 2) trailing and leading page buttons.

_First page_
first page

You can notice that there is no "First page" button, because page 1 button is already visible.

_Second page_
second page

Second page also does not have the "First page" button, for same reason as for page 1.

_Page N_
page n

When we get to page N, we can see trailing and leading N buttons (default 2), first page button (1), last page (10) and next and previous buttons. This way, user can proceed in reading posts chronologically, but he has an option to go to first and last page, as well as a way for "faster navigation" if a user is searing for some post.

_Last page_
last page

The "Last page" button will not be visible on last page. Simillar logic for the first page.

---

I hope all of this makes sense. You can clone my fork and try it out locally.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,6 +15,7 @@ destination: ./_site/blog
 markdown: kramdown
 permalink: /:title
 paginate: 8
+paginate_trail: 2
 paginate_path: /page:num/
 source: .
 excerpt_separator: "<!-- excerpt end -->"
--- a/_includes/pagination.html
+++ b/_includes/pagination.html
@@ -1,0 +1,42 @@
+{% if paginator.total_pages > 1 %}
+<div class="pagination d-none d-lg-block">
+    {% assign page_start = paginator.page | minus: site.paginate_trail %}
+    {% assign page_end = paginator.page | plus: site.paginate_trail %}
+
+    {% if paginator.previous_page == 1 %}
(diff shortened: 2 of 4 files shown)


tyhoff (approved):
Hey @MihajloNesic, thanks for the PR. I love it! I tried it out on the preview branch that Netlify publishes, and it works great! Merging!

And you are totally right...the pagination was getting out of hand. I think you came up with the right solution.

MihajloNesic:
Glad I can help :)

PR #204, opened 2021-06-11
post: example: mcuboot: freeze mcuboot to specific version

sophiekovalevsky:
This commit sets mcu boot repository used in example to a state where
the porting code implemented here is able to compile. The reason why
doing this is that, mcu boot so far has introduced several changes
that are missing in example. This will allow to compile the code right
away.


franc0is (changes requested):
Thank you for the PR, @sophiekovalevsky! One small thing: you forgot to refresh the submodules after checkout. Happy to merge once this is fixed!

sophiekovalevsky:
@franc0is thanks for that tiny detail. Should be ready now.

franc0is (approved):

franc0is:
Thank you! You are ok to merge.

PR #207, opened 2021-06-29
Prevent underflow by many backspaces in firmware-shell

Tagussan:
Successive backspace can break `rx_buffer` by making `rx_size` nagetive. I added if statement to prevent this.


franc0is (approved):
This is great, thank you @Tagussan !

PR #218, opened 2021-09-29
Article about CMSIS Packs

noahp:


noahp:
Ok, I think I got most of the typos fixed now :sweat_smile: Some awkward chunks of prose in there, overall structure might be a bit rough too, LMK!

PR #239, opened 2022-01-12
Add prettier action to pull requests

noahp:
### Summary

From https://github.com/memfault/memfault-docs .

Only files changed in the pull request will be formatted, see
[`.github/workflows/lint.yml`](.github/workflows/lint.yml).

Pull requests from branches on this repo will have an automatic commit applied
with the formatting. Pull requests from forks will **fail** CI, but not have a
commit applied.

 ### Test Plan

Test by touching a file under the following circumstances:

  - [x] the pull request is from a branch on this repo
    - https://github.com/memfault/interrupt/pull/243
    - the "Prettified code!" commit is auto-pushed to the branch
  - [x] the pull request is from a fork
    - https://github.com/memfault/interrupt/pull/244
    - the PR check fails: https://github.com/memfault/interrupt/runs/4938847953?check_suite_focus=true#step:5:19

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,0 +1,4 @@
+overrides:
+  - files: "*.md"
+    options:
+      proseWrap: always
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,60 @@
+name: Prettier
+
+on:
+  # Only run on pull requests, to avoid surprising commits when pushing to the
+  # main branch
+  pull_request:
+    branches: [master]
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.head.repo.full_name == github.repository
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Make sure the actual branch is checked out when running on pull requests
+          ref: ${{ github.head_ref }}
+
+      - id: files
+        uses: Ana06/get-changed-files@v2.1.0
+        with:
+          filter: |
+            *.js
+            *.md
+            *.mdx
+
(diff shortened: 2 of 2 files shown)


.prettierrc.yml
```diff
@@ -0,0 +1,4 @@
+overrides:
+  - files: "*.md"
+    options:
+      proseWrap: always
```
tyhoff, line 4: Are we sure we want this? It frequently mangles quote blocks and blocks of links. 

Have you had experiences with this?

tyhoff:
As much as I want this, I don't think it's ready for prime time. As a first example, it broke the lists in the IDAP post (first one ever!)

image
image

.prettierrc.yml
noahp: My experience has been pretty good with wrapping enabled, but it requires marking code blocks / quote blocks correctly :(

I'll change it to run only on changed files, and only on PRs (so we don't get a surprise on master).

noahp:
> As much as I want this, I don't think it's ready for prime time. As a first example, it broke the lists in the IDAP post (first one ever!)
> ...
> Or maybe we can comb through those issues. That issue above stems from the fact that the code block was not indented.

Very good call, I've updated it to only run on changed files and ONLY on PRs (so we don't get surprise commits). One thing to consider:

- currently this will auto-push to branches from this repo, and fail CI on PRs with branches from forked repos. this might be a bad/frictiony experience for users that have difficulty running `prettier`

noahp:
Closing, it's too much effort to get the skip behavior I want. The simple support github provides also disables the netlify preview action :confused: not ideal.

https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

PR #251, opened 2022-02-11
Fix unbalanced parens in noinit example code

mnaberez:
```
main.c:15:35: error: expected ')' before 'volatile'
   15 | __attribute__((section(".noinit")) volatile int my_non_initialized_integer;
      |                                   ^~~~~~~~~
      |                                   )
```
```
$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 10.3-2021.07) 10.3.1 20210621 (release)
```


chrisc11 (approved):
Thanks for the fix!

PR #254, opened 2022-02-28
Article on initializing padding in C structs

noahp:
There are a _ton_ of articles on this subject, but I couldn't find one
that gave practical examples and results for various zero-initialization
techniques, so I wrote one.

Also bonus section on a pending c23 improvement on this warty part of C!


_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+   };
+   ```
+
+3. use `{ 0 }` zero-initializer
+
+   ```c
+   struct foo a = { 0 };
+   ```
+
+4. use `{}` GCC extension zero-initializer
```
chrisc11, line 296: heh I was going to say list the doc to this extension but [apparently there is not one](https://gcc.gnu.org/legacy-ml/gcc/2019-07/msg00065.html). May be worth noting this syntax is valid C++ (just not C)

_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+It's tricky to recommend a one-size-fits-all option here, because different
+software will have different constraints. However, some general purpose advice
+follows.
+
+### Avoid relying on structure layout
+
+The simplest option to avoid padding issues is to avoid the padding fields
+altogether:
+
+> Access data directly via each member, do not alias structures or use `memcmp`
```
chrisc11, line 385: Maybe note it is okay to use memcmp if struct is packed though

_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+> etc.
+
+This only works if the padding data can be safely ignored in all use cases for
+the data structures in question.
+
+Another approach is to avoid structure holes entirely! The `pahole` tool could
```
chrisc11, line 391: 💯 I think in general one should strive for padding to only be present at end of struct :)

_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+```c
+// error prone! if the type of a changes, we might get unexpected results
+memset(&a, 0, sizeof(struct foo));
+
+// much better
+memset(&a, 0, sizeof(a));
+```
+
+### Last resort, `__attribute__((packed))`
+
```
chrisc11, line 416: I do think this should ALWAYS be used for data that may be read outside the application currently running whether that be another app or a future update of the same app where the data is reloaded from flash.

chrisc11 (approved):
💯 Really enjoyed this one!

_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+// Check 2 foos for equality
+bool foo_are_equal(struct foo a, struct foo b) {
+  return (a.i == b.i) && (a.b == b.b);
+}
+```
+
+If you are serializing the structure outside of your application (eg, over a
+UART, or writing it to non-volatile storage), but the data is _only_ ever read
+by your application, there's a good chance the padding bits won't matter.
```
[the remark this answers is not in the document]
noahp: good call, added a note

_drafts/c-struct-padding-initialization.md
noahp: good call, i was kinda stretching for anti-arguments but it really is best practice!

_drafts/c-struct-padding-initialization.md
noahp: added a callout for this

_drafts/c-struct-padding-initialization.md
```diff
@@ -0,0 +1,460 @@
...
+
+If on the other hand the structure needs to be parsed by another piece of
+software, it might be preferrable to pack the structure for easy to understand
+layout, at the expense of computation.
```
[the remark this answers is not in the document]
noahp: updated, good call :+1:

_drafts/c-struct-padding-initialization.md
noahp: sweet, added the reference! i can't even find documentation on this feature in the gcc sources (predates git!) :open_mouth: very secret corner of the language

noahp:
:heart: ty @chrisc11 for the excellent comments, applied them all!

franc0is (approved):

PR #260, opened 2022-03-31
Remove confusing file extensions in some snippets

jelledevleeschouwer:
@noahp, another great post! 

It seems that some file extension accidentally slipped into some of the snippets which led me to think I skipped a paragraph or something. But me thinks those are just artifacts from editing.


noahp (approved):
Whoops, thank you so much for fixing this! Much appreciated 🙏

PR #267, opened 2022-06-02
Renode peripheral name changed from uart2 to usart2

barneywilliams:
Renode changed the name of the UART peripherals from `uartx` to `usartx`, which breaks the Renode example. This PR changes `uart2` to `usart2`.


tyhoff (approved):
Thanks x2!

PR #270, opened 2022-06-21
Add some notes on `-ffile-prefix-map=old=new`

noahp:


_posts/2019-12-11-reproducible-firmware-builds.md
```diff
@@ -328,6 +328,28 @@ freertos_kernel/portable/MemMang/heap_1.c
...
+const char *my_file_name = "./test.c";
+```
+
+> Note: the `-ffile-prefix-map=old=new` was first added in GCC 8.1, although
+> apparently without enough excitement to deserve a mention in the [release
+> notes](https://gcc.gnu.org/gcc-8/changes.html)!<br/><br/>
+> This flag sets all `f*-prefix-map` settings, so it applies to debug info as
+> well as macro (`__FILE__`) substitutions. Note that it does NOT as of yet get
+> forwarded to `as` when invoked from the GCC frontend, so be sure to add
+> `-Wa,--debug-prefix-map=old=new` if that applies.<br/><br/>
```
chrisc11, line 347: Should this always be set? Would the previous instructions have had things covered?

chrisc11 (approved):

_posts/2019-12-11-reproducible-firmware-builds.md
noahp: Updated the note a smidge, it is probably a mistake in GCC (still required as of 12.1), since `-fdebug-prefix-map` does forward the arg to `as`.

PR #290, opened 2022-11-17
Quote cmsis pack manager path

dauc:
On my mac the cmsis-pack-manager path had a space in it.  `/Users/devin/Library/Application Support/cmsis-pack-manager`.


noahp (approved):
Amazing, thank you very much!

PR #292, opened 2022-12-13
Use lunr.js for the search box

noahp:
The Jekyll SimpleSearch does a pretty poor job searching article
titles/tags. Switch to lunr.js, which operates very similarly (all
client side, unlike eg angolia or google custom engine), but seems to do
a much better job, and even allows control over weighting!

image


noahp:
This does add about 10 seconds to live reload time on my machine, when lunr regenerates the search database 😕

victorbordo (approved):
❤️

_includes/search.html
```diff
@@ -1,23 +1,44 @@
 <!-- HTML elements for search -->
+
+<!-- This is installed from the https://github.com/slashdotdash/jekyll-lunr-js-search Gem -->
+<script src="/js/search.min.js" type="text/javascript" charset="utf-8"></script>
+
 <form id="search-form">
```
fnune, line 6: The `form` is currently making it so that hitting enter refreshes the page because the form gets submitted. I'd just remove it, since that's probably not what you want.

_includes/search.html
```diff
@@ -1,23 +1,44 @@
 <!-- HTML elements for search -->
+
+<!-- This is installed from the https://github.com/slashdotdash/jekyll-lunr-js-search Gem -->
+<script src="/js/search.min.js" type="text/javascript" charset="utf-8"></script>
```
fnune, line 4: This is a 138KB JS file. We should either put it right at the end of the body right before `</body>`.

If you do that, you should probably move the `$(...).lunrSearch` script to the bottom of the body also, because it depends on the 138KB bundle.

_includes/search.html
```diff
@@ -1,23 +1,44 @@
...
-    limit: 10,
-    fuzzy: false
-  })
+<script type="text/javascript">
+  $(function() {
+    $('#search-input').lunrSearch({
+      indexUrl  : '/js/index.json',           // url for the .json file containing search index data
+      results   : '#search-results',          // selector for containing search results element
+      template  : '#search-results-template', // selector for Mustache.js template
+      titleMsg  : '<h1>Search results<h1>',   // message attached in front of results (can be empty)
```
fnune, line 40: Would IMHO remove this since it's pretty obvious that these are the search results.

_includes/search.html
```diff
@@ -1,23 +1,44 @@
...
+<script src="/js/search.min.js" type="text/javascript" charset="utf-8"></script>
+
 <form id="search-form">
   <label for="search-input">
     <h3>Search</h3>
   </label>
-  <input type="text" id="search-input" placeholder="Search posts..">
+  <input type="text" id="search-input" name="q" placeholder="Search posts...">
 </form>
 <ul id="results-container"></ul>
```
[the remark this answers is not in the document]
fnune: (looks like this was the container for the previous search)

_includes/search.html
```diff
@@ -1,23 +1,44 @@
...
+
+<!-- This is installed from the https://github.com/slashdotdash/jekyll-lunr-js-search Gem -->
+<script src="/js/search.min.js" type="text/javascript" charset="utf-8"></script>
+
 <form id="search-form">
   <label for="search-input">
     <h3>Search</h3>
   </label>
-  <input type="text" id="search-input" placeholder="Search posts..">
+  <input type="text" id="search-input" name="q" placeholder="Search posts...">
```
fnune, line 10:
```suggestion
  <input autocomplete="off" type="text" id="search-input" name="q" placeholder="Search posts...">
```

Getting the autocomplete dropdown from the browser was a bit annoying to me. This should disable it.

fnune (approved):
Left some tips. I also like that you use `article`, but I think the results look too large and the tags in a `ul` make the list look inconsistent (some posts don't have tags).

I reworked it a bit. Feel free to discard it if you don't like it!

image

Here's the whole `search.html` with the changes I made. The CSS is terrible, sorry. Also, note that (as recommended in the other comments) I removed the scripts from this file, they should go right before the closing `</body>` tag.

```html
<form id="search-form">
  <label for="search-input">
    <h3>Search</h3>
  </label>
  <input type="text" id="search-input" name="q" placeholder="Search posts...">
</form>

<section id="search-results" style="display: none;"> </section>

{% raw %}
<script id="search-results-template" type="text/mustache">
  {{#entries}}
    <article class="search-result">
      <h1><a href="{{url}}">{{title}}</a></h1>
      <small>
        {{#date}}<time datetime="{{pubdate}}" pubdate>{{displaydate}}</time>{{/date}}
        {{#is_post}}
          {{#tags}} #{{.}}{{/tags}}
        {{/is_post}}
      </small>
    </article>
  {{/entries}}
</script>
{% endraw %}

<style>
  #search-results > h1 {
    display: none;
  }

  #search-results {
    padding-top: 1.3em;
  }

  .search-result {
    padding-bottom: 0.5em;
  }

  .search-result h1 {
    font-size: 1rem;
    line-height: 1.5;
  }
</style>
```

_includes/search.html
fnune: (by "this" I mean the `titleMsg`)

I realized that removing this line actually doesn't remove the `h1`... so I added some dirty CSS to hide it in my suggestion in the main comment.

_includes/search.html
fnune: Note that the autocomplete from the browser for this search input should be pretty useless...

tyhoff:
As for the Mac side - I ran into 2 issues on my Mac machine that I'll need to push fixes for. I have them stashed away locally - was waiting for the frontend bits to be reworked before pushing forward because that was the part that I wasn't going to be able to fix. 

One is that libv8 doesn't work, so we have to use the master branch of the search plugin (which I forked and is now in our repo) and the other fix is around recursion limits in JSON parsing.

_includes/search.html
fnune: Another way about this if you don't wanna fiddle with CSS after you remove the form is to add this attribute to it... :grimacing: 

```
action="javascript:void(0);"
```

_includes/search.html
noahp: i set it to `''` which did the necessary!

_includes/search.html
noahp: done

_includes/search.html
noahp: done!

noahp:
> As for the Mac side - I ran into 2 issues on my Mac machine that I'll need to push fixes for. I have them stashed away locally - was waiting for the frontend bits to be reworked before pushing forward because that was the part that I wasn't going to be able to fix.
> 
> One is that libv8 doesn't work, so we have to use the master branch of the search plugin (which I forked and is now in our repo) and the other fix is around recursion limits in JSON parsing.

@tyhoff have at it, I'm done tinkering w/ the front end stuff!

noahp:
I updated the Docker image here to be compatible with this change:
#294

chrisc11 (approved):
neat!

PR #296, opened 2023-01-10
Load the lunr search script on all pages

noahp:
It was only loaded on the main page- add it to all pages that include
the `script.html` fragment.


_includes/script.html
```diff
@@ -3,3 +3,18 @@
 <!-- Start of HubSpot Embed Code -->
 <script type="text/javascript" id="hs-script-loader" async defer src="//js.hs-scripts.com/8551838.js"></script>
 <!-- End of HubSpot Embed Code -->
+
+<!-- This is installed from the https://github.com/slashdotdash/jekyll-lunr-js-search Gem -->
+<script src="{{ '/js/search.min.js' | relative_url }}" type="text/javascript" charset="utf-8"></script>
```
fnune, line 8: Why `relative_url`? I'd have assumed it's always `/js/search.min.js`.

fnune (approved):
Assuming `_includes/script.html` is included at the bottom of the HTML :+1:

_includes/script.html
noahp: ah yep, copy-paste error! ty :heart:

tyhoff (approved):

PR #300, opened 2023-01-30
Draft: Coroutines article

koniarik:
As discussed in email, I am making a PR with a draft of article about coroutines, I am looking forward for cooperation!

P.S: the article is intentionally set to 2024, feel free to suggest better date.


_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+#### Dynamic memory
+
+The first thing that I assume most of you noticed at the beginning is the fact the coroutine requires dynamic memory.
+This is not favourable in embedded in many cases, but there are ways to avoid that.
+
+Coroutines have `allocator` support, we can provide the coroutine with an allocator that can be used to get memory for the coroutine and hence avoid the dynamic allocation. (Approach that I can suggest)
+This is done by implementing custom `operator new` and `operator delete` on the `promise_type` which allocates the `entire frame`.
+
+An alternative is to rely on `halo` optimization (Heap Allocation Elision Optimization) if the coroutine is implemented correctly and the parent function executes the entire coroutine in its context.
+The compiler can optimize away the dynamic allocation and just store the frame on the stack of the coroutine's parent.
```
bahildebrand, line 210: To me this seems cleaner than a custom allocator. Could you elaborate on why you prefer the custom allocator solution?

_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+This can be enforced by deleting appropriate `operator new` and `operator delete` overloads of the `promise_type`, but it seems clumsy to me.
+
+And to kinda ruin the pretty thing here, there is one catch. As of now, the compiler can't tell you how much memory the coroutine needs, as it is known only during the link time - the size of the coroutine frame can't be compiler constant. (TODO: link to the source for this)
+This means that you effectively can't prepare static buffers for the coroutines.
+
+What I would suggest (that is what I do), is to use coroutines for the long-term process and just build them during the initialization of the device.
+Or just live with the dynamic memory.
+
+(Note: if I could dream a bit here, I would want an explicit way of forcing the coroutine to live on the stack, which would have clearly defined behaviour of when and how it should happen, and with proper compiler errors if one fails to do so)
```
bahildebrand, line 219: Yeah, this would be the preferred way :(. It might help to flesh out the `halo` optimization more and why it isn't as ideal, since by your description it looks to accomplish this.

_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+    std::span<const uint8_t> data = co_await i2c_receive_awaiter(addr, 42);
+    // do something with data
+}
+```
+
+This has the benefit that while the `i2c operation` is processed by the peripheral, the main loop can be busy doing something else, and the execution is not stuck in the `interact_with_device` with busy waiting.
+And we do not have to pay for that by having complexity (manual state machine doing the interaction) or having threads (which bring in other problems).
+
+Given the bus nature of `i2c`, it is also quite easy to achieve sharing of the `i2c bus` with multiple `device drivers` for various devices on the bus itself.
+We can just implement the `i2c_coroutine round_robin_run(std::span<i2c_coroutine> coros)` coroutine that uses round robin to share access to the peripheral between multiple coroutines (devices).
```
bahildebrand, line 286: I think this section would be very interesting to elaborate. Other strategies for handling concurrency for this scenario, and how you would implement them.

bahildebrand (commented):
Overall I like the idea of the article. I think our readers would be interested in a more focused look at how to implement coroutines for asynchronous peripheral IO, and what the benefits over traditional threaded concurrency would look like. Maybe you could spin up an asynchronous driver for an MCU peripheral and use it as an example? 

Let me know if you have any question about any of my feedback.

_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+ 1) implement it as an iterative process and schedule it
+ 2) use threads for the processes
+
+Let's take communication over UART as an example:
+ 1) iterative process means the usage of structure to store the state of the communication, and provide a function that based on the current state advances the state. (Commonly with the usage of an enum representing actual status)
+ 2) implement the communication in a thread which might release its time if it waits for some operation
+
+Approach 1) has multiple potential issues, it might take a lot of code to implement a complex exchange of data in this way (a lot of state variables) and the approach is problematic, as there is a chance that one of the steps might take longer than expected and we can't prevent that.
+
+Approach 2) has another set of issues. Each thread requires its own stack space (which might not scale), and we got all the problems of parallelism - exchanging data between can suffer various potential concurrency issues.
```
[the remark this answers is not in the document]
koniarik: Eeeeeh, it does, point of coroutines is that you make a context swithc at specific points in the code that are controlled by you. While in normal freertos-like threading, the context switch can happen anytime the main timer fires and switches the threads.

This means that you do not have to think problems about stuff like strict atomic work with shared variables, let's assume that i is shared variable between parts of the system that execute concurrently. This is unsafe in threads, but is perfectly ok in case of coroutines:

```cpp
int j = i;
j += 1;
i = j;
```

That is, with threads: context switch can happen at any point in time, at any instruction, and you DO have to think about it and take care
with coros: context switch happens ONLY when you explicitly write in the code that the coroutine should interrupt itself.

For me this is one of the biggest motivations to use coroutines over threads in general, as the mental overhead of managing concurrent access to stuff with this explicit cooperation is much smaller than with threads.

Anyway, I will try to re-think this part and think how to express it in better way :)

_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+
+There are multiple potential issues with coroutines and using them correctly, but I think there are a few that are noteworthy for the embedded community.
+
+#### Dynamic memory
+
+The first thing that I assume most of you noticed at the beginning is the fact the coroutine requires dynamic memory.
+This is not favourable in embedded in many cases, but there are ways to avoid that.
+
+Coroutines have `allocator` support, we can provide the coroutine with an allocator that can be used to get memory for the coroutine and hence avoid the dynamic allocation. (Approach that I can suggest)
+This is done by implementing custom `operator new` and `operator delete` on the `promise_type` which allocates the `entire frame`.
```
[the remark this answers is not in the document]
koniarik: Understood, I will think about expanding this part more clearly, generally: you can use whatever you want to get/free memory, that is: malloc/free, your own memroy resource, allocators or anything.

_posts/2024-04-20-coroutines.md
koniarik: I will ok this more to make sure it is explained properly, `halo` is optimization... you can't force compiler to do it, comiler usually won't tell you why it did it or did not do it, or it won't even tell you how exactly it decides

_posts/2024-04-20-coroutines.md
```diff
@@ -0,0 +1,348 @@
...
+This means that you effectively can't prepare static buffers for the coroutines.
+
+What I would suggest (that is what I do), is to use coroutines for the long-term process and just build them during the initialization of the device.
+Or just live with the dynamic memory.
+
+(Note: if I could dream a bit here, I would want an explicit way of forcing the coroutine to live on the stack, which would have clearly defined behaviour of when and how it should happen, and with proper compiler errors if one fails to do so)
+
+#### Frame size is big
+
+One more invisible caveat that appeared is that GCC is not yet smart enough with the frame size.
```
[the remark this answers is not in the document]
koniarik: I will try to check this

_posts/2024-04-20-coroutines.md
koniarik: I added some round robin implementation, wil lthink about others

PR #317, opened 2023-05-09
Fix annoying message when search index is missing

noahp:
In #315 I made the lunr.js search index generation opt-in via the
environment variable `GENERATE_SEARCH_INDEX`, but did not add a check
for the missing index file in that case, and that results in an error
when the page is loaded:

```
[2023-05-09 16:20:42] ERROR `/js/search.min.js' not found.
```

Update our forked plugin and the scripts template to omit the script for
non-production build; switch to the built-in environment variable
`JEKYLL_ENV` which is available to Liquid template expressions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ gem 'jekyll-seo-tag', "~> 2.7.1"
 
 group :jekyll_plugins do
   gem 'jekyll-diagrams'
-  gem 'jekyll-lunr-js-search', :git => 'https://github.com/memfault/jekyll-lunr-js-search.git', :ref => '4db4cab461594a92022430fd75442747494f4160'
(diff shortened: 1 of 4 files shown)


fnune (approved):
I still use the indexed search in localhost and it works. Won't this disable it for anything other than `docs.memfault.com`?

Approving anyway, no big deal.

noahp:
> I still use the indexed search in localhost and it works. Won't this disable it for anything other than `docs.memfault.com`?
> 
> Approving anyway, no big deal.

oh yeah, let me add more flavor. in #315 I made the index gen opt-in, because usually locally you only care about a single page anyway, and the search gen takes some seconds (~10+) which makes live reload a little sad

PR #323, opened 2023-05-24
Amend README to remove the --incremental option when running locally

ejohnso49:
This flag causes jekyll to not regenerate the page index which causes new posts to not show up in the blog list.

After further investigation, this behavior is known and documented: https://jekyllrb.com/docs/configuration/incremental-regeneration/

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -30,7 +30,7 @@ $ git clone https://github.com/memfault/interrupt.git
 $ cd interrupt
 $ pip install -r requirements.txt
 $ bundle install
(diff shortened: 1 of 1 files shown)


noahp:
How's the performance without this flag?

ejohnso49:
> How's the performance without this flag?

Adds a few seconds to a hot-reload. The thing that I think kills using `--incremental` is that the blog index is only updated after running `clean` and then building which negates any benefit. It's impossible to experiment with new pages quickly using this option

noahp (approved):
Nice find, tyvm!

PR #333, opened 2023-06-30
The post about how the interrupt controller may be used as a simple scheduler

romanf-dev:


_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+On initialization phase, each actor subscribes to some queue. Once message arrives to that queue the actor is activated. Since each actor may be subscribed to exactly one queue, subscription may be implemented as return value: the actor should return a valid pointer to the queue it wants to be subscribed to. Return values may change during an actor’s lifetime but must always be a valid queue. This is because of the absence of internal queue: when an actor has not been subscribed to a queue, it will have no chance to run again. So, pseudocode for an actor is shown below:
+
+
+        actor(self, message) {
+            ... message handling ...
+            free(message)
+            return queue_to_be_subscribed_to
+        }
```
bahildebrand, line 79: I think it would be nice to have a working code sample for this. Could you add something to the `example` directory?

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+Queues may contain both messages and actors. If someone posts a message in an empty queue, it will save that message until some actor will try to poll. On the other hand, if an actor tries to get a message from an empty queue it will be saved as a subscriber and later when a message arrives the actor will be activated. Therefore, a queue at any moment of time may contain either messages or actors, but never both. Note that queue is just a head of linked list, it has no internal buffers, so queue overflow cannot happen.
+
+For simplicity, messages are fixed-sized. Message allocator is just a pool of fixed-sized memory chunks. For latency-critical systems, it may be even implemented as lock-free. When the message pointer arrives to an actor, it is assumed that message type is known to its recipient since it knows what queue it polls.
```
bahildebrand, line 84: Would it make more sense for this to just be a fixed size queue? Then we could avoid allocations. Or is that what you're getting at?

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+During this process, the system remains fully preemptable and asynchronous: other interrupt handlers may post messages to other queues, if corresponding actors are delayed then messages just will be accumulated in queues. When some actor or interrupt activates another high-priority actor, then preemption occurs immediately, so this system has good response times and less jitter than loop-based solutions.
+
+Normally, the actor model does not require explicit synchronization like semaphores and mutexes because there are no mutable shared state. However, there may be shared resources, such as devices. In such a case mutual exclusion may be represented as multiple actors posting their requests to a single queue and one actor associated with that queue who will do all the work sequentially. Nevertheless, in practice, it is often desirable to block preemption. This may be done using hardware features for temporarily raising a running handler’s priority to some ceiling. Ability to block certain interrupt priorities without disabling interrupts completely also helps with response times since high-priority actors may never be blocked.
```
bahildebrand, line 95: I'm not sure I follow why we'd want to be in a critical section when in these actor interrupts. Wouldn't we want that so higher priority actors could take over if they have work to do?

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+{% include newsletter.html %}
+
+{% include submit-pr.html %}
+<!-- Interrupt Keep END -->
+
+{:.no_toc}
+
+## References
+
+- The actor model in 10 minutes:
```
bahildebrand, line 140: You can link these directly where they're referenced by adding `[^1-N]` here and in the spots they're referenced they will link automatically. The template has an example of how that should be set up.

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+
+In general, the actor model is a bad choice when you need some kind of global consensus or transactions (i.e., rollback of previous work when some actor at the end of the chain fails). So, while not all systems are suitable for the actor model, it may be a good choice for some and hardware implementation of parts of the model in widely available MCUs makes it yet more attractive.
+
```
bahildebrand, line 129: It would be nice to have `Final Thoughts` section or something similar summarizing what you covered in the article.

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+In general, the actor model is a bad choice when you need some kind of global consensus or transactions (i.e., rollback of previous work when some actor at the end of the chain fails). So, while not all systems are suitable for the actor model, it may be a good choice for some and hardware implementation of parts of the model in widely available MCUs makes it yet more attractive.
```
bahildebrand, line 128: We generally like to have a stronger recommendation for what is being written about in the closing section. Try to give a compelling argument as to why what you're proposing is good, or has the ability to be good after some drawbacks are solved.

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+
+Normally, the actor model does not require explicit synchronization like semaphores and mutexes because there are no mutable shared state. However, there may be shared resources, such as devices. In such a case mutual exclusion may be represented as multiple actors posting their requests to a single queue and one actor associated with that queue who will do all the work sequentially. Nevertheless, in practice, it is often desirable to block preemption. This may be done using hardware features for temporarily raising a running handler’s priority to some ceiling. Ability to block certain interrupt priorities without disabling interrupts completely also helps with response times since high-priority actors may never be blocked.
+
+Note that the interrupt vector is assigned to priority level, not to an actor, so the total number of actors is unlimited:
+
+![](/img/hw_actors/vectors.png)
```
bahildebrand, line 99: I'm not sure I follow the design entirely. So the run queues have multiple actors in them, and those actors themselves have message queues to receive messages?

bahildebrand (commented):
Interesting article! I left some comments for some areas I think need clarification. Also could you run your article through the free version of grammarly? Usually we just paste it in and apply some fixes. From there we can do another rev on structuring/grammar.

_drafts/hw_actors.md
romanf-dev: Hi, thank you for the feedback! I will try to fix all the issues you mentioned. Since fixing requires that parts of the article have to be rewritten it may take some time. So, I will update the PR once fixes will be ready. I have already used analog of grammarly as stated in submission manual (I experience issues with the grammarly itself) but may be this is still not enough.

_drafts/hw_actors.md
```diff
@@ -0,0 +1,150 @@
...
+
+The approach is to use this model and map priorities to interrupt vectors with user-defined priority. Note that this model contains no threads and runs at a single shared stack. It is inherently asynchronous and even applicable to multiprocessor systems since (in the case of GIC) you may trigger interrupts on other CPUs too.
+
+
+## The concept
+
+
+Since the actor model itself is a very complex topic for a single pocket article, instead of talking about abstractions, I mostly focus on their representation in the MCU environment.
+
+In the classic actor model, an actor is just a message handler similar to interrupt handlers. Unlike threads, it has no dedicated stack and its purpose is to react to messages arriving into its internal mailbox.
```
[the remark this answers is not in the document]
romanf-dev: Yes, they all share the interrupt stack since they are essentially run-to-completion routines. I mean there they do not need separate stack as threads do.

_drafts/hw_actors.md
romanf-dev: You’re right, it’s a little confusing. I mean short critical sections to achieve atomicity. For example, if we want something like debug output to a UART, it is a bad idea to use device registers in concurrent environment without any synchronization, so we want to block preemption until output is done. I removed this paragraph in the updated version because its is possibly misleading.

_drafts/hw_actors.md
romanf-dev: When allocation happens inside the queue implementation it would require copying user-provided message payload into the internal storage. While it may be OK I mostly focus on speed and simplicity in this article.

_drafts/hw_actors.md
romanf-dev: Actors have no internal storage (except the pointer to the message they’re handling), queues are separate objects that are not shown. Initially each actor is subscribed to some queue, since they are waiting they are not in the runqueues. Once someone pushes a message to the queue, the subscribed actor is moved to runqueue. The picture depicts a situation where multiple actors with different priorities are ready to run and therefore multiple vectors are pending. I hope I fixed it in the updated version.

gminn (commented):
Hi Raman! Neat article, I enjoyed reviewing it. I have a couple comments. Let me know if you have any questions.

_drafts/hw_actors.md
```diff
@@ -0,0 +1,173 @@
...
+
+If you look at the programmer's model of interrupt controllers like [NVIC](https://developer.arm.com/documentation/ddi0337/e/Nested-Vectored-Interrupt-Controller?lang=en) or GIC, you will notice that these devices do exactly that. For example, in the NVIC case, the device contains a set of registers describing the state of interrupt vectors, their priorities and so on. Also, it has a special register called STIR (Software Trigger Interrupt Register). Writing a vector number to that register causes the corresponding interrupt to be activated in the same way as hardware does. Since the user can also change the priority of interrupts, the interrupt controller itself may be treated as a simple scheduler or an implementation of the so-called [actor model](https://en.wikipedia.org/wiki/Actor_model) in **hardware**.
```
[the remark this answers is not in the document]
romanf-dev: Thank you for the review. All minor issues and suggestions are fixed in the next commit.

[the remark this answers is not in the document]
gminn: Great, thank you!

bahildebrand (approved):
LGTM!

PR #334, opened 2023-06-30
Create June 2023 Roundup

geekigirlx:
- Noah's video on Raspberry Pi; in Articles & Learning, but should I create a "Random" section and add this there? Or would it make sense to add it to News & Announcements if it's an update on the shortage? 
- @bahildebrand saw your reply to my email about Mohammed's article. I'll add that next. Would this description work? "Memfault Ambassador Mohammed Billoo (MAB Labs) demystifies the “Kconfig” infrastructure that is an integral part of The Zephyr Project RTOS" -- borrowed directly from the newsletter
- Tyler's share on Bristlemouth; I think this is in the right spot (Projects & Tools) but LMK if I'm off
- Chris's share on the STM32U5 series; I left out the top line of what he shared on slack (https://theinterrupt.slack.com/archives/C01B3FET876/p1686936713910099). Does it make sense as I have it, or do I need to include "^ The devkit makes use of ST's STM32U5 series."? The way it was phrased/located made it a bit confusing.

### DIFF
--- a/_posts/2023-07-05-june-2023-roundup.md
+++ b/_posts/2023-07-05-june-2023-roundup.md
@@ -1,0 +1,83 @@
+---
+title: "What we've been reading in June"
+author: tyler
+tags: [roundup]
+---
+
+<!-- excerpt start -->
+
+Here are the articles, videos, and tools that we've been excited about this
+June. 
+
+<!-- excerpt end -->
+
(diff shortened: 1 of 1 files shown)


bahildebrand (approved):
Looks good Matti! If you want to add Mohamed's article we should be good to go.

PR #355, opened 2023-08-02
content: update about summary to feature DRE

bahildebrand:


noahp (approved):
could use some more emojis :smiling_face_with_tear: (jk looks great!)

PR #361, opened 2023-08-10
 missing robots.txt in Interrupt fixed

bkmxer:


nihalgonsalves:
Thanks @bkmxer! For the robots.txt this should be enough:

```
Sitemap: https://interrupt.memfault.com/sitemap.xml

User-agent: *
Allow: /
```

The disallows shouldn't reference anything in the _repository_ like a gitignore,  just what's deployed. I don't think we have any paths we need to disallow.

I also believe the sitemap should be `https` because we serve over https only (http is 301 redirected)

nihalgonsalves (approved):

PR #362, opened 2023-08-10
Add github action to check for redirects on rename

noahp:
See it catch the error here:

https://github.com/memfault/interrupt/actions/runs/5825914834/job/15798644392#step:4:9

See it pass when a redirect is included here:

https://github.com/memfault/interrupt/actions/runs/5826059210/job/15799126175#step:4:5

### DIFF
--- a/.github/workflows/detect-renamed-posts.js
+++ b/.github/workflows/detect-renamed-posts.js
@@ -1,0 +1,54 @@
+const fs = require("fs");
+const path = require("path");
+const child_process = require("child_process");
+
+const redirectsFilePath = path.join(
+  path.dirname(path.dirname(__dirname)),
+  "_redirects"
+);
+
+function updateRedirects(originalName, newName) {
+  const redirects = fs.readFileSync(redirectsFilePath, "utf-8");
+  const lines = redirects.split("\n");
+
+  const newRedirect = `/blog/${originalName} /blog/${newName}`;
+
+  // search for a line with newRedirect, if not found, raise an error
+  const found = lines.find((line) => line.trim() === newRedirect.trim());
+  if (found) {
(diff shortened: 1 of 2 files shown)


.github/workflows/detect-renamed-posts.js
```diff
@@ -0,0 +1,54 @@
...
+const gitCommand = `git diff --name-status --diff-filter=R -M5 HEAD^ -- _posts/`;
+const renamedFiles = child_process
+  .execSync(gitCommand)
+  .toString()
+  .split("\n")
+  .filter((file) => file !== "");
+renamedFiles.forEach((file) => {
+  // each line should have 3 elements:
+  // R100    _posts/2020-01-01-old-post.md    _posts/2020-01-01-new-post.md
+  const fileParts = file.split("\t");
```
bahildebrand, line 44: My only realy concern is here. I think this is reasonable, but probably not really a contract from the git cli that these will always be tabs. That being said I don't have a better suggestion, so I think I'm just being :old_man_shakes_fist_at_javascript:

bahildebrand (approved):

.github/workflows/detect-renamed-posts.js
noahp: :grimacing: oh yeah, this thing is SUPER brittle, there is undoubtedly much better libraries/options out there

gminn (approved):
This is great. Nice add!

.github/workflows/detect-renamed-posts.js
bahildebrand: Eh, I say we send it. If it breaks it can always be disabled. I doubt this will change much in the future

PR #364, opened 2023-08-13
Add a link to the sandbox in sidebar

bkmxer:
Additionally, added a notification banner template as an example for future reference.


gminn (approved):

_includes/share-buttons.html
```diff
@@ -26,6 +26,8 @@ <h3 class="">Embedded Events</h3>
     </p>
 </div>
 
+{% include notification-banner.html %}
```
gminn, line 29: Given that this is no longer a configurable notification banner and rather a sidebar item, it would be best to incorporate the html directly here to match the other sidebar items.

_data/_notifications-template.yml
```diff
@@ -0,0 +1,6 @@
+# Notification Banner example
+Instrumenting Sentry:
+    background: '#141845'
+    textcolor: '#fff'
+    title: Instrumenting Sentry
+    description: Instrumenting 1 Sentry for your backend project? Join us July 11th, at 10 AM PT for the Backend Error Monitoring 101 livestream. <a href="https://go.memfault.com/how-to-debug-android-devices-production">Register now.</a>
```
gminn, line 6: This template will be nice for the future when we may need it -- thank you for adding it! Only note - could you add a new line at the end of this file?

gminn (changes requested):
Thank you for adding this! Excited to have the sandbox callout. I have a few requests before we merge.

_includes/share-buttons.html
bkmxer: Hey @gminn 
So just to make sure, we aren't looking to make it configurable any more, just plain HTML, right?

gminn (approved):

gminn:
Thank you! I adjusted the PR title and description to reflect the shift to just the sandbox update in sidebar, and the example notification banner.

PR #370, opened 2023-08-24
Helpful note for running scan-build

TimTTP:
Hi, this blog post was really useful but missing this information took me way too long to research! I have seen this issue with clang-10 (as used in the post) and clang-16 (latest). This issue is mentioned in the comments of the post and the `gcc` default on non-Darwin platforms is in the docs (https://clang-analyzer.llvm.org/scan-build.html) if you search for "Darwin".

This would have saved me a lot of time and presumably will help people in the future!


noahp (approved):
Thank you!

PR #377, opened 2023-09-12
content: publish cortex-m interrupt actors article

bahildebrand:


_drafts/hw_actors.md
```diff
@@ -1,18 +1,21 @@
...
+Many embedded RTOS now have at least a few thousand lines of code, so a simple, lightweight framework is often needed for embedded devices. Since ARM is the most popular embedded CPU and almost every ARM processor has a hardware scheduler, it would be interesting to make a compact framework utilizing these features. The most popular approach to RTOS-less event-driven systems is super-loop consuming events produced by interrupts. This article discusses an alternative, more efficient approach in some circumstances. It has less overhead than RTOS and also benefits event-driven systems.
```
bahildebrand, line 10: I'm still not fully sold on mentioning the number of lines of code as an indicator of complexity. Maybe we could rephrase this along the lines of:

```
Using an RTOS is often a tradeoff between the ease of decomposing tasks, with the complexity of the scheduler itself. There exists a middle ground between highly complex systems that may require an RTOS, and simpler ones that can be easily modeled using a super loop.
```

romanf-dev:
I think everything is ok except the following minor issue:

“ For example, NVIC spec is fixed, but several available priorities and vectors may vary from one vendor to another.”

The vendor is free to define the whole number of available vectors and priorities. It may be misleading, which exact vectors are mentioned as “several”. Perhaps it would be better as:

“ For example, NVIC spec is fixed, but the count of available priorities and vectors may vary from one vendor to another.”

romanf-dev (approved):

_drafts/hw_actors.md
```diff
@@ -23,28 +26,26 @@ This is why concepts where interrupts are used as the main 'execution engine' ar
 
 ## The hardware scheduler
```
gminn, line 27: Nit: add capitalization for each heading

_drafts/hw_actors.md
```diff
@@ -23,28 +26,26 @@ This is why concepts where interrupts are used as the main 'execution engine' ar
...
+If you look at the programmer's model of interrupt controllers like [NVIC](https://developer.arm.com/documentation/ddi0337/e/Nested-Vectored-Interrupt-Controller?lang=en) or GIC, you will notice that these devices do exactly that. For example, in the NVIC case, the device contains a set of registers describing the state of interrupt vectors, their priorities, and so on. Also, it has a special register called STIR (Software Trigger Interrupt Register). Writing a vector number to that register causes the corresponding interrupt to be activated in the same way as hardware does. Since the user can also change the priority of interrupts, the interrupt controller itself may be treated as a simple scheduler or an implementation of the so-called [actor model](https://en.wikipedia.org/wiki/Actor_model) in a **hardware** peripheral.
 
 
 ## The concept
 
-
 The classic actor model describes an actor as a message handler, similar to an interrupt handler. When handling messages, the actor may:
 
 - create new actors
```
gminn, line 38: Not as important but a nit would be capitalizing the first work of each bullet here too

romanf-dev (approved):

_drafts/hw_actors.md
```diff
@@ -23,28 +26,26 @@ This is why concepts where interrupts are used as the main 'execution engine' ar
...
+All available interrupt vectors are divided into two classes: those producing and consuming messages. Interrupts designated to peripheral devices are expected to post messages to queues that, in turn, have subscribed actors.
 
-Let's implement simplest 'blinky' application with one actor who handles messages from systick interrupt and controls onboard LED. The full source of the example for STM32 Bluepill board may be found [here](/example/hw-actors/main.c). The messages have to be sent from systick interrupt to the actor controlling the LED. Therefore, message payload is the next LED state:
+Let's implement the simplest 'blinky' application with one actor who handles messages from systick interrupt and controls onboard LED. The whole source of the example for the STM32 Bluepill board may be found [here](/example/hw-actors/main.c). The messages must be sent from systick interrupt to the actor controlling the LED. Therefore, the message payload is the next LED state:
 
         ```c
```
gminn, line 50: Looks like markdown doesn't like the indenting here so the syntax highlighting isn't coming through. Just need to left-justify and it works 👍

PR #378, opened 2023-09-15
Fix broken link to source code

gminn:
### Summary

A relative path used to point to source code on GitHub worked when navigating
from GitHub, but doesn't work when accessed from a different domain name, i.e.
memfault.com. The link is now the full GitHub URL.

### Test Plan

Tested manually with local build.


biggerfisch (approved):
LGTM!

PR #379, opened 2023-09-15
printf article

noahp:


.gitignore
```diff
@@ -8,3 +8,4 @@ _site/
 /vendor
 /example/*/build
 /.venv
+/node_modules
```
bahildebrand, line 11: Why was this suddenly needed?

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+
+## `printf`
+
+Ah, the first and last of debugging techniques!
+
+Let's first explore what happens when printf runs. Here's a simplified flowchart
+showing what happens when `printf` is called:
+
+{% blockdiag size:120x40 %} blockdiag { A [label = "printf"]; B [label =
+"vprintf"]; C [label = "_write syscall"]; A -> B -> C; } {% endblockdiag %}
```
bahildebrand, line 40: Didn't know you could do this. Fancy

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+  // return the built in global reentrancy context
+  return _impure_ptr;
+}
+```
+
+Also be sure to also define `__DYNAMIC_REENT__` in your CFLAGS when building
+your application source.
+
+This will trip a software breakpoint if a non-reentrant function is called from
+an interrupt context!
```
bahildebrand, line 192: I'm not sure I follow the value of this. Not sure if it's my brain being smooth, but why would we want to trigger a software breakpoint here? Would this just be for local debugging to figure out if we were using these functions in a re-entrant way?

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+    default:
+      // crash
+      __asm__("bkpt 0");
+  }
+
+  return isr_impure_ptr;
+}
+```
+
+Note that if you're using an RTOS, this may already be handled in some way; or
```
bahildebrand, line 265: Zephyr works like this no? Worth a direct callout?

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+which might be needed for your actual application.
+
+Also, configuring a UART might require managing clock configuration, chip
+power/sleep modes, figuring out buffering the data (DMA?), making sure
+everything is interrupt safe, etc.
+
+And finally, you'll usually need some kind of adapter dongle to interface the
+UART with your PC (eg USB-to-TTL or USB-to-asynchronous serial).
+
+But! because the interface is contained entirely on the microcontroller, it can
```
bahildebrand, line 345: Exclamation point is kind of weird, but I won't protest if you want to leave it for flavor

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+monitor semihosting ioclient 2
+```
+
+(From <https://wiki.segger.com/J-Link_GDB_Server>).
+
+For pyocd, you can pass the `--semihosting` argument when starting the gdb
+server.
+
+As an extra bonus, see below for a simplified semihosting solution based on the
+newlib one that doesn't require linking against `rdimon.specs`, and saves some
```
bahildebrand, line 410: Not sure if I missed, but earlier it was mentioned that we would provide more info on rdimon. Might be good to provide an overview or link before this

_drafts/printf-on-embedded.md
```diff
@@ -0,0 +1,566 @@
...
+
+<!-- Interrupt Keep END -->
+
+{:.no_toc}
+
+## References
+
+<!-- prettier-ignore-start -->
+
+- [Interrupt: Introduction to ARM Semihosting]({% post_url 2021-02-16-arm-semihosting %})
```
bahildebrand, line 549: Can we add these links directly to the spots they're referenced? Might mike it easier for people to dive deeper as they read.

bahildebrand (commented):
Looks great overall! I still need to do a grammar pass through, but just had a few comments otherwise.

.gitignore
noahp: ah yeah, i had added some mermaid node stuff for diagramming but pulled it out :facepalm: will revert! tyty

_drafts/printf-on-embedded.md
noahp: oh yeah good call, will tidy this up- the idea is to crash if using a non-reentrant function from isr

_drafts/printf-on-embedded.md
noahp: they're all just copied from above, all the links are inline :grimacing: personally I prefer having bare urls, do we prefer footnote links or hyperlinks?

_drafts/printf-on-embedded.md
noahp: done!

_drafts/printf-on-embedded.md
bahildebrand: I don't think we have a hard preference. I like the superscript links personally, but I think it's fine as is.

bahildebrand (approved):
Nice job! Feel like I learned more than I ever thought I could have about printf!

PR #387, opened 2023-10-12
Add webinar recording & URL + Sandbox demo 

geekigirlx:
Added blockquote for battery metrics webinar recording with link and added blockquote for sandbox demo under the Memfault section of the article.

### DIFF
--- a/_posts/2023-07-26-monitoring-battery-life.md
+++ b/_posts/2023-07-26-monitoring-battery-life.md
@@ -17,6 +17,11 @@ In this post, I’ll cover how to start thinking about collecting metrics that c
 
 <!-- excerpt end -->
 
+> 📺 **Watch the Webinar Recording**
+>
+> Do you tend to favor video content over written materials? This topic was also the subject of a [Memfault Webinar: An Empowering Endeavor - Predicting and Improving Device Battery Life](https://hubs.la/Q023ppsY0).
+
+
 {% include newsletter.html %}
 
 {% include toc.html %}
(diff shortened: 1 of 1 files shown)


_posts/2023-07-26-monitoring-battery-life.md
```diff
@@ -481,6 +486,10 @@ All of us at Memfault have thought about how to monitor battery life quite exten
 
 To learn more about how you might instrument battery life with Memfault, which is quite similar to this post, check out the Memfault Documentation page, [Tracking Battery Life with Memfault](https://docs.memfault.com/docs/best-practices/metrics-for-battery-life/).
 
+> **Try the Sandbox**
+>
+> Want to get hands-on with Memfault? With the Sandbox, you can interact freely with a bunch of real data, or if you prefer, follow one of our guided product tours. No need to integrate any Memfault SDK - [test the platform on your own immediately](https://hubs.la/Q025ldvt0).
+
```
bahildebrand, line 492: This seems kind of out of place here. Should it be after the conclusion?

geekigirlx: Ah, I thought it made sense to include it in the Memfault section. Would it be better if I changed the wording to nod to monitoring battery life with Memfault, as that section suggests? Or can move to end, but whenever we can place a call to action earlier on (as opposed to the end), and so long as it makes sense, that would be the preference. That said - if it doesn't make sense lmk.

bahildebrand: Yeah, it's the right spot. I just can't read.

bahildebrand (approved):

PR #407, opened 2023-12-13
publish: lte and iot article

gminn:


tyhoff (commented):
- [We stopped calling them pocket articles](https://memfault.slack.com/archives/CJ7ARQHF1/p1682619849923529) - the name wasn't doing them justice. 
- I think you need to clarify your goal with the article. If the goal is to give a history of all cellular data protocols, 

I think you should add some personal story. Introduction:

> I started at Memfault a few months back, and the first major project I started on was to dig into LTE metrics on the Nordic Thingy:91. Honestly, I knew very little about LTE, but I'm happy to say I've come out the other end of the project having learned more about it than I ever imagined. We also have [link to PR] new metrics going up into the Nordic SDK soon, and I now carry two Thingy91's in my backpack to help test Memfault and improve our LTE analyses.
> 
> In this article, ....

EDIT - Looks like I read your first introduction and didn't realize you had another introduction down below. I would try to combine the two, and use the excerpt to pull out one of the paragraphs.

- Split up the large paragraphs before the section "Cells, UEs, and eNBs". Some are quite hefty.
-

tyhoff (approved):

PR #428, opened 2024-02-11
Update GitHub action versions in C Development Environment article

lmapii:
This is just a little update for the GitHub actions mentioned in the article. I updated the reference repository already, and those should be the matching changes in the article.


noahp:
Nice fix, thank you! I keep running over old repos of mine with the same problem 😅

noahp (approved):

PR #433, opened 2024-02-28
Upgrade to Ruby 3

tyhoff:
Use 3.2.2 which is already installed on Cloudflare Pages


tyhoff:
Won't do this - too difficult as we'll need to update or remove lunar-js search functionality.

PR #443, opened 2024-03-21
Create 2024-03-21-diving-into-jtag-part-3.md

geekigirlx:
Three things I want to call out: 

1. There is a table in a codeblock that was rendered quite nicely in Medium (which puts "function" on row one, column four in purple text - not possible to do in markdown, I've checked) - so I did my best to mimic that. Not an exact replica but close. The alternative would be to use a screenshot instead.

2. Images. I know we want to save the images in a folder on Interrupt. 4 of the figures/images shown are actually animated GIFs. I have saved all the images locally and can create a folder to share, but I'm unsure of how to add these to the Interrupt images folder, especially the animated ones. Will the way I've shown the images work (which use the image address URL?). 

3. Lastly - at the very top/beginning, I changed the article description from what the previous JTAG (Part 2) had. Can't remember if they all had the same description, can go back and look but I don't believe so.

### DIFF
--- a/_posts/2024-03-21-diving-into-jtag-part-3.md
+++ b/_posts/2024-03-21-diving-into-jtag-part-3.md
@@ -1,0 +1,216 @@
+---
+
+title: Diving into JTAG — Debugging (Part 3)
+description:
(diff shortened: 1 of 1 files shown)


gminn:
Draft was merged in another PR

PR #413, opened 2023-12-18
Draft: Practical Zephyr part 6 "West workspaces"

lmapii:
Hi, this PR contains the sixth and final draft article for the article series about Zephyr. This one's about west workspaces.

PR #454, opened 2024-04-10
Draft: Practical Zephyr part 6 "West workspaces"

lmapii:
Rebase of https://github.com/memfault/interrupt/pull/413/

### DIFF
--- a/_drafts/practical_zephyr_05_west.md
+++ b/_drafts/practical_zephyr_05_west.md
@@ -1,0 +1,1246 @@
+---
+title: Practical Zephyr - West workspaces (Part 6)
+description: Article series "Practical Zephyr" the sixth part about West workspaces.
+author: lampacher
+tags: [zephyr, west]
+---
+
+<!-- excerpt start -->
+
+In the previous articles, we used _freestanding_ applications and relied on a global Zephyr installation. In this article, we'll see how we can use _West_ to resolve global dependencies by using _workspace_ applications. We first explore _West_ without even including _Zephyr_ and then recreate the modified **Blinky** application from the previous article in a _West workspace_.
+
+<!-- excerpt end -->
+
+After this article, you should know _West_ not as a meta-tool for `build` and `flash` commands, but as a powerful dependency manager for your Zephyr applications. Hopefully - after reading this article - you will no longer create _freestanding_ Zephyr applications but will use _West workspaces_ only.
+
+{% include newsletter.html %}
+
+{% include toc.html %}
+
+## Prerequisites
+
+This article is part of the _Practical Zephyr_ article series. In case you haven't read the previous articles, please go ahead and have a look. This article requires that you're able to build and flash a Zephyr application to the board of your choice and that you're familiar with _Devicetree_.
+
+We'll be using the [development kit for the nRF52840](https://www.nordicsemi.com/Products/Development-hardware/nrf52840-dk) and the [STM32 Nucleo-64 development board](https://www.st.com/en/evaluation-tools/nucleo-f411re.html), but you can follow along with any target - real or virtual.
+
+> **Note:** This article uses two _manifest repositories_ that are available on GitHub:
+> - [practical-zephyr-t2-empty-ws](https://github.com/lmapii/practical-zephyr-t2-empty-ws) contains an empty _manifest_ repository for creating a dummy _West_ workspace that we'll use to explore _West_ without Zephyr.
+> - [practical-zephyr-manifest-repository](https://github.com/lmapii/practical-zephyr-manifest-repository) contains the _manifest repository_ of the Zephyr application that we're building in this article.
+
+> **Note:** While writing this article series, I had to switch from STM's Nucleo-C031C6 to the Nucleo-F411RE since I broke my C031C6 development kit. Apologies for the subtle switch - but you should be able to follow along with any board.
+
+
+
+## Creating a workspace from scratch
+
+Let's start from scratch, without any tools installed. Ok, that might be a bit much - let's say you have at least [`git`](https://git-scm.com/), [`python`](https://www.python.org/) and [`pipenv`](https://pipenv.pypa.io) installed. You can, of course, use `pyenv` or install directly using `pip`, whatever works best for you!
+
+```bash
+$ mkdir workspace
+$ cd workspace
+```
+
+
+### Installing west
+
+We've been using [West](https://docs.zephyrproject.org/latest/develop/west/index.html) a lot now to build, debug and flash our project. It comes bundled with the [nRF Connect SDK](https://www.nordicsemi.com/Products/Development-software/nrf-connect-sdk), so until now, we didn't have to install it at all. Without sourcing any paths, e.g., from a `setup.sh` script that we created at the very beginning of this article series, I don't have _West_ installed:
+
+```bash
+$ west --version
+zsh: command not found: west
+```
+
+The first thing to even get started is therefore installing West, which is available as a `python` package. Instead of following the [official documentation](https://docs.zephyrproject.org/latest/develop/west/install.html) and installing _West_ globally, I'll be using `pipenv` to manage an environment for me:
+
+```bash
+$ pwd
+$ path/to/workspace
+$ touch Pipfile
+```
+
+In my `Pipfile`, I can simply add `west` as a `dev-package`, since I won't be needing it for any `python` application or extension:
+
+```ini
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[dev-packages]
+autopep8 = "*"
+pylint = "*"
+west = "1.2"
+
+[requires]
+python_version = "3.11"
+```
+
+All that's left to do is to install the environment and activate the shell, and we have _West_ available:
+
+```bash
+workspace $ pip install pipenv
+workspace $ pipenv install --dev
+workspace $ pipenv shell
+(workspace) workspace $ west --version
+West version: v1.2.0
+```
+
+In the remainder of this article, I won't explicitly refer to the `python` environment or _West_ installation anymore. It is up to you to decide which approach to use, e.g., installing _West_ globally, using [`pipenv`](https://pipenv.pypa.io/), [`venv`](https://docs.python.org/3/library/venv.html), [`poetry`](https://python-poetry.org/), or maybe you're using one Zephyr's [docker images](https://github.com/zephyrproject-rtos/docker-image) - whatever works best for you!
+
+
+### Adding a manifest
+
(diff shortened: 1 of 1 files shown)


_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V] <command> ...
+west: unknown command "build"; workspace /path/to/workspace-m does not define this extension command -- try "west help"
+```
+
+As we've seen in previous articles, e.g., `build` and `flash` are Zephyr-specific [extension commands](https://docs.zephyrproject.org/latest/develop/west/zephyr-cmds.html) that are not included by a plain _West_ installation. We'll resolve this in the following sections.
+
+
+### Setting up the manifest repository and file
+
+For our workspace structure, we're using what's known as a ["T2 star topology"](https://docs.zephyrproject.org/latest/develop/west/workspaces.html#topologies-supported): In such a topology, the _manifest repository_ does not only contain the _manifest file(s)_ but also contains all application files. If you'd move your application code to a separate repository, populated in turn again using a "project" in the manifest file, you'd end up with Zephyr's documentation calls a "T3: forest topology".
```
ejohnso49, line 588: I think I would strike the sentence on the T3 topology since you cover this later on

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+  self:
+    path: app
+
+  projects:
+    - name: zephyr
+      revision: v3.4.0
+      url: https://github.com/zephyrproject-rtos/zephyr
+      path: deps/zephyr
+      # add Zephyr-specific West extensions
+      west-commands: scripts/west-commands.yml
```
ejohnso49, line 697: I might recommend skipping this for imports

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+```
+
+> **Notice:** Zephyr provides [environment scripts](https://docs.zephyrproject.org/latest/develop/env_vars.html#zephyr-environment-scripts) including a `zephyr-env.sh`, which you can source in case you're using [Zephyr's official SDK](https://docs.zephyrproject.org/latest/develop/toolchains/zephyr_sdk.html).
+
+In this script, I'm extending my `$PATH` for the binaries that come with Nordic's toolchain installation. This includes `cmake`, `ninja`, and all other host tools used by Zephyr. In fact, this installation even contains `git` and `python`, but by appending to `$PATH`, my local installations have precedence over Nordic's binaries.
+
+Then, I'm setting the `ZEPHYR_TOOLCHAIN_VARIANT` to `zephyr` and point `ZEPHYR_SDK_INSTALL_DIR` to the full path of the toolchain installation. Since I'm still using my own virtual environment, I need to install all Python packages required by Zephyr. Instead of cloning Zephyr as documented in the [getting started guide](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#install-dependencies), I can use the locally populated Zephyr installation:
```
ejohnso49, line 779: It's not clear to me what last sentence here means. I might leave it out instead, it's clear to me that the next step is to install the Python dependencies

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+    path: app
+
+  projects:
+    - name: zephyr
+      revision: v3.4.0
+      url: https://github.com/zephyrproject-rtos/zephyr
+      # the path is no longer needed since we're now using `path-prefix`
+      # path: deps/zephyr
+      # add Zephyr-specific West extensions
+      west-commands: scripts/west-commands.yml
```
ejohnso49, line 846: I think you could also note that `west-commands` is no longer needed with the import as well

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+      # recursively import Zephyr dependencies
+      import:
+        path-prefix: deps
+        file: west.yml
```
ejohnso49, line 850: Might be good to note that this is not strictly needed as this is the default value

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+manifest:
+  projects:
+    - name: zephyr
+      revision: v3.5.0
+      url: https://github.com/zephyrproject-rtos/zephyr
+      west-commands: scripts/west-commands.yml
+      import: true
+```
+
+These dependencies are something that you need to handle yourself since [_West_ ignores projects that have already been defined in other files](https://docs.zephyrproject.org/latest/develop/west/manifest.html#manifest-imports). E.g., if you define _Zephyr_ as a project in your manifest file, and _after_ it add [Golioth's Zephyr SDK](https://docs.golioth.io/firmware/zephyr-device-sdk/), then the _Zephyr_ project in the `west-zephyr.yml` file of the Golioth SDK is _ignored_.
```
ejohnso49, line 1040:
```suggestion
These dependencies are something that you need to handle yourself since [_West_ ignores projects that have already been defined in other files](https://docs.zephyrproject.org/latest/develop/west/manifest.html#manifest-imports). E.g., if you define _Zephyr_ as a project in your manifest file, and _after_ add [Golioth's Zephyr SDK](https://docs.golioth.io/firmware/zephyr-device-sdk/), then the _Zephyr_ project in the `west-zephyr.yml` file of the Golioth SDK is _ignored_.
```

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+ * branch            main       -> FETCH_HEAD
+error: Your local changes to the following files would be overwritten by checkout:
+        test.txt
+Please commit your changes or stash them before you switch branches.
+Aborting
+...
+ERROR: update failed for project dummy
+```
+
+The only irreversible mistake that we can make is to commit changes to the `manifest-rev` branch. Let's do just that:
```
ejohnso49, line 1146: From the example here it looks like the commit is made from a detached head, not from the `manifest-rev`

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+If you want to keep it by creating a new branch, this may be a good time
+to do so with:
+
+ git branch <new-branch-name> 8de0853
+
+HEAD is now at d22a413 updated test.txt
+...
+```
+
+The output that you see above is just Git's output. You can't actually restore the commit `8de0853` since West discarded it, the changes are really lost.
```
ejohnso49, line 1180: I don't think this is right, things should be recoverable in this state. The commit will not be garbage collected immediately so the output recommending `git branch <new-branch-name> 8de0853` should work.

For the other case of committing to the `manifest-rev` branch, things are trickier but not impossible to recover. In my testing, I see west move the branch to the destination commit, but the changes to the previous state are staged allowing you to recover from there.

ejohnso49 (commented):
Looks great, had a few things that I think need to be edited

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+├── Pipfile
+└── Pipfile.lock
+```
+
+
+### Locally vs. remotely initialized workspaces
+
+What is the difference between a _West workspace_ that has been initialized locally using the `--local` flag, or remotely by passing the URL of the manifest repository? As mentioned already, thankfully, the short answer is _none_.
+
+The only difference is that for remotely initialized workspaces _West_ clones the repository for you and thus you essentially don't need to use `git clone` for _any_ of the repositories in your workspace: West takes care of this, "forget" about `git clone` if you like.
```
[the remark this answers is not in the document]
lmapii: Yep, you're right. I mixed up cloning the manifest repo with cloning the dependencies!

_drafts/practical_zephyr_05_west.md
```diff
@@ -0,0 +1,1246 @@
...
+workspace $ west update
+workspace $ # nothing happens
+```
+
+Huh, that was disappointing. `west update` didn't do anything - feel free to check your files, nothing changed! The reason for that is quite simple: With `west init` we already took care of initializing the workspace. Since we don't have any external dependencies to manage, we have no work for `west update`.
+
+
+### Adding projects
+
+In your manifest file, you can use the `manifest.projects` key to define all Git repositories that you want to have in your workspace. Let's say that we're planning to use a more modern [doxygen](https://doxygen.nl/) documentation for our project and therefore want to add [Doxygen Awesome](https://github.com/jothepro/doxygen-awesome-css) to our workspace. We can add this external dependency as an entry in the `manifest.projects` key of our manifest file:
```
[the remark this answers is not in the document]
lmapii: I wish it was Doxygen's default theme :D

_drafts/practical_zephyr_05_west.md
lmapii: You mean you'd not add `west-commands`? Without `import` at this point it'd be necessary to keep it since otherwise the `west` extension commands would not work, but I'll add a note.

_drafts/practical_zephyr_05_west.md
lmapii: In the getting started guide they instruct you to use `west init ~/zephyrproject` to get the Zephyr sources and to then install all dependencies using `pip install -r ~/zephyrproject/zephyr/scripts/requirements.txt`. I didn't find this an adequate solution since any Workspace application anyhow has Zephyr as a project.

Would you have a suggestion how to phrase this instead?

_drafts/practical_zephyr_05_west.md
lmapii: Yep, I'll fix this.

_drafts/practical_zephyr_05_west.md
lmapii: Yes, but that's the actual output of the command line. That was actually the point I was trying to make later in the article: The Git output of West can be very misleading.

_drafts/practical_zephyr_05_west.md
lmapii: Yep, I didn't try this! Seems to work, I'll update the article. Thanks!

lmapii:
Thanks a lot for the review! I hope I fixed things accordingly. On a sidenote I still can't see the deployment preview. Could it be that this is no longer working for pull-requests from forks?

_drafts/practical_zephyr_05_west.md
ejohnso49: Oh right, good point. I think leaving things here is fine since we have the later comment

ejohnso49 (approved):

PR #463, opened 2024-04-25
fix: rename file extensions in ew post

gminn:
### Summary

Cloudflare deployment preview is case-insensitive while
the prod Cloudflare deployment is not it seems! On the
[current post](https://interrupt.memfault.com/blog/embedded-world-2024), only the first image renders because it has
`.jpg` while the others have `.JPG`.

Also recalled through this that the MacOS filesystem is
case-insensitive so had to use `git mv` to get git to
recognize the filename changes 🥲


bahildebrand (approved):
:exploding_head:

PR #481, opened 2024-06-13
lint: add lefthook, prettier, typos

noahp:
Install from npm, because it's simpler, so use lefthook as the
pre-commit runner (`yarn install` automatically installs the hooks,
which is somewhat nice too).

I removed the reference to the `grammarly` vscode extension, since it's
no longer available in the marketplace and the github repo is archived.


package.json
```diff
@@ -0,0 +1,14 @@
...
+    "lint:prettier": "prettier --check _posts _drafts",
+    "lint:typos": "typos _posts _drafts",
+    "lint:prettier:fix": "prettier --write  _posts _drafts"
+  },
+  "dependencies": {
+    "lefthook": "^1.6.16",
+    "prettier": "^3.3.2"
+  },
+  "devDependencies": {
+    "typos-rs-npm": "dalisoft/typos-rs-npm"
```
gminn, line 12: Suggestion - you could add a empty `_typos.toml` file to anticipate us needing to add false positives (I just ran it on my LTE article and UE wasn't recognized as a valid word). I'm good with adding it when we need it though!

readme.md
```diff
@@ -36,9 +41,13 @@ You can now access the server at [http://0.0.0.0:4000](http://0.0.0.0:4000)
 ### Locally
 
 You'll need:
+
 - Python 3.8 or later
 - Ruby 2.7.7
 
+Optionally, to run the linters, you'll need `yarn` (easy way is via
```
gminn, line 48: I'd say we take out optional  and say its required so we can have everyone running these checks

gminn (approved):
Thank you so much for adding this! been wanting it for a while ❤️  Can't wait to open the massive formatting and typos fixing PR 💥

package.json
noahp: good call, done!

gminn (approved):

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# See https://bun.sh/docs/install/lockfile
+*.lockb binary diff=lockb
```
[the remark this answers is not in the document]
gminn: I think this review comment makes sense

gminn:
Got this branch building again 💪  I also confirmed that the lefthook `prettier` and `typos` checks are working well locally! 

The error from build logs was:
```
➤ YN0070: Migrating from Yarn 1; automatically enabling the compatibility node-modules linker 👍
```

Changes I made to fix it:

- Use Yarn 3:
```
yarn set version 3.6.3
```
- Add `nodeLinker: node-modules` to `.yarnrc.yml` to configure yarn 3.

- Finally, reset env and regenerate `yarn.lock`:

```
rm -rf node_modules && rm yarn.lock
yarn install
```

PR #486, opened 2024-06-27
Update ruby

noahp:
Old ruby doesn't work on my ubuntu machine default version.

Just update. We're so out of date!

I confirmed the block diagrams render correctly still:

https://noahp-update-ruby.interrupt.pages.dev/blog/device-firmware-update-cookbook#dfu-code-should-be-updatable

image


requirements.txt
```diff
@@ -1,3 +1,4 @@
-blockdiag
-Pillow==9.5.0
+# Install a newer version of blockdiag, that's compatible with modern pillow
+git+https://github.com/blockdiag/blockdiag.git@refs/pull/179/head
```
gminn, line 2: Frustrating that they haven't merged this yet

.ruby-version
```diff
@@ -1 +1 @@
-2.7.7
\ No newline at end of file
+3.2.2
```
gminn, line 1: Was there a particular reason to go to 3.2.2 instead of latest (3.3.3)?

gminn (approved):
😮‍💨 2.7 is [already EOL](https://www.ruby-lang.org/en/downloads/branches/). thanks for upgrading!

.ruby-version
noahp: unfortunately the cloudflare builder is on 3.2.2, and updating it in `cloudflare.sh` didn't seem to work :/

cloudflare_build.sh
```diff
@@ -1 +1,2 @@
-bundle exec jekyll build --config _config.yml${JEKYLL_BUILD_CONFIG_PARAM} && cp _redirects _site/_redirects
\ No newline at end of file
+asdf plugin update ruby
```
noahp, line 1: revert!

.ruby-version
gminn: ah makes sense. sad

PR #490, opened 2024-07-10
Revert "Update ruby (#486)"

gminn:
This reverts commit ec597d6419ae039e13dca64dc75f08e7f6959988.

PR #493, opened 2024-07-17
Upgrade ruby to 3.2.2, replace lunr.js with algolia search

noahp:
- **Reapply "Update ruby (#486)" (#490)**
- **algolia**
- **temporarily use the index for docs.memfault.com for testing, until we get the interrupt index set up**


Gemfile.lock
```diff
@@ -49,62 +38,60 @@ GEM
...
+    strscan (3.1.0)
     terminal-table (2.0.0)
       unicode-display_width (~> 1.1, >= 1.1.1)
     unicode-display_width (1.8.0)
-    webrick (1.7.0)
+    webrick (1.8.1)
 
 PLATFORMS
-  ruby
+  x86_64-linux-gnu
```
gminn, line 83: Is there a reason you added the linux platform? Seeing that there may be a concern with hosting services requiring the Linux platform to be specified https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/#platforms. Wondering if you ran into this?

_includes/script.html
```diff
@@ -4,19 +4,13 @@
...
-        emptyMsg  : '<p>Nothing found.</p>'     // shown message if search returns no results
-      });
-    });
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
+<script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3" ></script>
+<script>
+docsearch({
+  container: '#docsearch',
+  appId: '4SXF6GYFKZ',
+  apiKey: '2028276a4fc7f09eeb0bd082fbd85164',
```
gminn, line 13: Should we rotate this and put it in a github secret, or is it not very sensitive?

gminn (changes requested):
A couple questions! The preview looks great 🤩 🔍

Gemfile.lock
noahp: Whoops, this is an error- my bundler generates platform-specific lockfile by default, need to fix it!

_includes/script.html
noahp: It's exposed in the frontend code in the generated site, not considered secret- it's a read-only key.

gminn (approved):
Looks great! All good to go once we get the interrupt index

noahp:
#528 replaces this, closing

PR #515, opened 2024-09-14
Draft/wsl2 firmware development

JPHutchins:
Hi!  I'm a long time reader, first time contributor.

This is a pair of articles about the use of WSL2 for firmware development, with Zephyr as a convenient example.

One articles assesses the viability - pros and cons - of using WSL2 for FW development.  The other article is a tutorial for setting up WSL2 for FW development.

Before the articles are published, the links from one to the other will need to be updated since they currently link to the article location in the _drafts folder.

Mermaid support is added, which is optionally enabled with `mermaid: true` at the top of an article.

Cheers,
JP


JPHutchins:
Rebased and split up into #516 and #517.

PR #522, opened 2024-10-04
Add a devcontainer config file

noahp:
This enables using the VSCode devcontainer extension.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,22 @@
+{
+    "name": "Interrupt",
+    "build": {
+        "dockerfile": "../Dockerfile"
+    },
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "esbenp.prettier-vscode",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "esbenp.prettier-vscode",
+                "DavidAnson.vscode-markdownlint"
+            ]
+        }
+    },
+    "postAttachCommand": ". /venv/bin/activate && bundle exec jekyll serve --host 0.0.0.0 -D --trace --future",
```
tyhoff, line 14: Can we just have it use `entrypoint.sh` so there aren't two copies of this command?

noahp: ah yep lemme test, good idea

noahp: yep it's good! ty sir

PR #560, opened 2025-03-06
Create 2025-03-06-february-2025-roundup.md

geekigirlx:
@gminn can you double-check my descriptions that are owned by François and make sure they sound decent? 

Also, does that callout under the Interrupt live section look okay? Is it just me or is the font gray (and if so is that fine?)? 

On that note, how do we feel about a separate Interrupt Live section? Otherwise we can next those under events.

### DIFF
--- a/_drafts/2025-03-06-february-2025-roundup.md
+++ b/_drafts/2025-03-06-february-2025-roundup.md
@@ -1,0 +1,90 @@
+---
+title: "What we've been reading in February (2025)"
+description:
+  Here are the articles, videos, and tools we've been excited about in February
+  2025.
+author: gminn
+tags: [roundup]
+---
+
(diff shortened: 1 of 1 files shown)


gminn (approved):
> can you double-check my descriptions that are owned by François and make sure they sound decent?
✅ 
> does that callout under the Interrupt live section look okay?
✅ 

>Is it just me or is the font gray (and if so is that fine?
That's completely fine, that is the Markdown syntax highlighting being a bit confusing

>how do we feel about a separate Interrupt Live section?
Looks great

gminn:
### Merge activity

* **Mar 7, 8:58 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/memfault/interrupt/560).

PR #631, opened 2026-03-27
noahp/gatsby

noahp:
- **Move dark mode toggle to hamburger menu**
- **Port to gatsby**


noahp:
not really important right now

